PR #1, opened 2016-10-18
Utillity for working with Base64 in JSON

NegativeMjark:


base64.go
```diff
@@ -0,0 +1,30 @@
...
+	return json.Marshal(base64.RawStdEncoding.EncodeToString(b64))
+}
+
+// UnmarshalJSON decodes a JSON string and then decodes the resulting base64.
+func (b64 *Base64String) UnmarshalJSON(raw []byte) (err error) {
+	// We could add a fast path that used base64.RawStdEncoding.Decode
+	// directly on the raw JSON if the JSON didn't contain any escapes.
+	var str string
+	err = json.Unmarshal(raw, &str)
+	if err != nil {
```
Kegsay: Nit: Can in-line this.

base64.go
```diff
@@ -0,0 +1,29 @@
...
+// MarshalJSON encodes the bytes as base64 and then encodes the base64 as a JSON string.
+func (b64 Base64String) MarshalJSON() ([]byte, error) {
+	// This could be made more efficient by using base64.RawStdEncoding.Encode
+	// to write the base64 directly to the JSON. We don't need to JSON escape
+	// any of the characters used in base64.
+	return json.Marshal(base64.RawStdEncoding.EncodeToString(b64))
+}
+
+// UnmarshalJSON decodes a JSON string and then decodes the resulting base64.
+func (b64 *Base64String) UnmarshalJSON(raw []byte) (err error) {
```
Kegsay: Here you're using a pointer receiver but `MarshalJSON` uses a value receiver. Intentional?

Kegsay: c.f - https://golang.org/doc/faq#methods_on_values_or_pointers

> Next is consistency. If some of the methods of the type must have pointer receivers, the rest should too, so the method set is consistent regardless of how the type is used.

Kegsay:
LGTM

PR #2, opened 2016-10-18
Add function for looking up matrix servers in DNS

NegativeMjark:


resolve.go
```diff
@@ -0,0 +1,108 @@
...
+	SRVError   *DNSError             // If there was an error getting the SRV records.
+	Hosts      map[string]HostResult // The results of looking up the SRV record targets.
+	Addrs      []string              // List of "<ip>:<port>" strings that the server is listening on.
+}
+
+// A DNSError describes an error that occured when trying to lookup a matrix server in DNS.
+type DNSError struct {
+	Error     string // The string decription of the error.
+	Temporary bool   // Whether the error is temporary.
+	Timeout   bool   // Whether the error was a timeout.
```
Kegsay: I feel that we are being misleading here by using `bool` and not `*bool`. If the error isn't a `net.Error` then both `Temporary` and `Timeout` will be `false`, which feels wrong. I feel like we should say "we dunno" rather than mislead the caller like this.

resolve.go
```diff
@@ -0,0 +1,108 @@
...
+	result.Error = err.Error()
+	if neterr, ok := err.(net.Error); ok {
+		result.Temporary = neterr.Temporary()
+		result.Timeout = neterr.Timeout()
+	}
+	return
+}
+
+// LookupServer looks up a matrix server in DNS.
+func LookupServer(serverName string) (result DNSResult, err error) {
```
Kegsay: A 60 line function is waaaaay longer than I'd like if you're using named return args. Please, think of my sanity!

resolve.go
```diff
@@ -0,0 +1,108 @@
...
+
+		if err != nil {
+			if dnserr, ok := err.(*net.DNSError); ok {
+				if dnserr.Timeout() {
+					return
+				}
+				hosts[serverName] = []net.SRV{net.SRV{
+					Target: serverName,
+					Port:   8448,
+				}}
```
Kegsay: A comment explaining that we can recover by assuming the default Matrix port of 8448 would be great.

Kegsay: Also, won't this be completely ruined by the fact that `err` is non-nil at this point? So:
```go
_, err := LookupServer("matrix.org"); err != nil {
  // will execute
}
```

resolve.go
```diff
@@ -0,0 +1,108 @@
...
+		addrs, err := net.LookupHost(cname)
+		result.Hosts[host] = HostResult{
+			CName: cname,
+			Addrs: addrs,
+			Error: dnsError(err),
+		}
+		for _, record := range records {
+			for _, addr := range addrs {
+				ipPort := net.JoinHostPort(addr, strconv.Itoa(int(record.Port)))
+				result.Addrs = append(result.Addrs, ipPort)
```
Kegsay: Is it really more useful to have addresses as `matrix.org:8448` and not split out into host + port fields? I genuinely don't know how this API will be consumed, hence asking.

resolve.go
NegativeMjark: Fixed 7adbd0a

resolve.go
NegativeMjark: Fixed 7adbd0a

resolve.go
```diff
@@ -0,0 +1,108 @@
...
+// A DNSResult is the result of looking up a matrix server in DNS.
+type DNSResult struct {
+	SRVCName   string                // The canonical name for the SRV record in DNS
+	SRVRecords []*net.SRV            // List of SRV record for the matrix server.
+	SRVError   *DNSError             // If there was an error getting the SRV records.
+	Hosts      map[string]HostResult // The results of looking up the SRV record targets.
+	Addrs      []string              // List of "<ip>:<port>" strings that the server is listening on.
+}
+
+// A DNSError describes an error that occured when trying to lookup a matrix server in DNS.
```
NegativeMjark: Fixed 7adbd0a

NegativeMjark:
@Kegsay I've removed the DNSError object, and pass the raw network error instead.

If the client wants to inspect it then it can cast the error to a net.Error itself.

Kegsay:
LGTM

PR #3, opened 2016-10-18
Add utility functions for encoding canonical JSON

NegativeMjark:


json.go
```diff
@@ -0,0 +1,186 @@
...
+import (
+	"encoding/binary"
+	"encoding/json"
+	"sort"
+	"unicode/utf8"
+)
+
+// CanonicalJSON re-encodes the JSON in a cannonical encoding. The encoding is
+// the shortest possible encoding using integer values with sorted object keys.
+func CanonicalJSON(input []byte) ([]byte, error) {
```
Kegsay: Do we have a spec for this? Can you link to it in the comments please? E.g. http://matrix.org/docs/spec/server_server/unstable.html#canonical-json ?

json.go
```diff
@@ -0,0 +1,186 @@
...
+		firstChar = input[i]
+		i++
+	}
+	if firstChar == '[' {
+		return sortJSONArray(input, output)
+	}
+	if firstChar == '{' {
+		return sortJSONObject(input, output)
+	}
+	return append(output, input...), nil
```
Kegsay: You should make it clear in the docs that if you don't give `SortJSON` valid JSON it will just spit back the input you gave *rather than return an error*.

Kegsay: I suspect you're doing this so you can recursively call this?

json.go
```diff
@@ -0,0 +1,186 @@
...
+	sep := byte('[')
+	for _, value := range array {
+		output = append(output, sep)
+		sep = ','
+		if output, err = SortJSON(value, output); err != nil {
+			return nil, err
+		}
+	}
+	if sep == '[' {
+		output = append(output, '[', ']')
```
Kegsay: Gnarly. An inline comment to explain why you need to do this would be good.

json_test.go
```diff
@@ -0,0 +1,75 @@
...
+	if err != nil {
+		t.Error(err)
+	}
+	if string(got) != want {
+		t.Errorf("CompactJSON(%q): want %q got %q", input, want, got)
+	}
+}
+
+func TestSortJSON(t *testing.T) {
+	testSortJSON(t, `[{"b":"two","a":1}]`, `[{"a":1,"b":"two"}]`)
```
Kegsay: Cheeky tests to avoid having to check for spaces, since your sort is also removing whitespace. Mentioning that somewhere (like the `SortJSON` docs...) would be nice.

json.go
```diff
@@ -0,0 +1,186 @@
...
+				if c == '"' {
+					break
+				}
+			}
+		}
+	}
+	return output
+}
+
+func compactUnicodeEscape(input, output []byte, index int) ([]byte, int) {
```
Kegsay: Not a single comment exists in this function, do you *really* think that's appropriate?

json.go
```diff
@@ -0,0 +1,186 @@
...
+		var buffer [4]byte
+		n := utf8.EncodeRune(buffer[:], rune(codepoint))
+		output = append(output, buffer[:n]...)
+	}
+	return output, index
+}
+
+func readHexDigits(input []byte) uint32 {
+	hex := binary.BigEndian.Uint32(input)
+	// substract '0'
```
Kegsay: `substract`

Kegsay:
Much much better, thank you! The only outstanding problem I see now is the fact that `SortJSON` is misleading. That function will also partly compact the input bytes because we skip over whitespace. This should be mentioned in the docs given it is a public function and that there exists a specific `CompactJSON` which makes the caller think that *that* is the thing which strips whitespace. 

LGTM

PR #4, opened 2016-10-19
Add functions for adding and verifying JSON signatures

NegativeMjark:


signing.go
```diff
@@ -0,0 +1,105 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"golang.org/x/crypto/ed25519"
+)
+
+// SignJSON signs a JSON object returning a copy signed with the given key.
+// https://matrix.org/docs/spec/server_server/unstable.html#signing-json
+func SignJSON(entityName, keyID string, privateKey ed25519.PrivateKey, message []byte) ([]byte, error) {
```
Kegsay: Please stick to the same terminology for variable names as is used in the spec. That is:
 - `s/keyID/signingKeyID/`
 - `s/entityName/signingName/`

signing_test.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+}
+
+type MyMessage struct {
+	Unsigned   *json.RawMessage `json:"unsigned"`
+	Content    *json.RawMessage `json:"content"`
+	Signatures *json.RawMessage `json:"signature,omitempty"`
+}
+
+func TestSignJSONWithUnsigned(t *testing.T) {
```
Kegsay: I would like to see some `TestVerifyJSON` tests which is expected to fail due to:
 - No signatures
 - No signature for key ID
 - Nonsense signature

Given at present your tests will succeed if:
```go
func VerifyJSON(entityName, keyID string, publicKey ed25519.PublicKey, message []byte) error {
  return nil
}
```

signing_test.go
```diff
@@ -0,0 +1,77 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	signed, err := SignJSON(entityName, keyID, privateKey, input)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	err = VerifyJSON(entityName, keyID, publicKey, signed)
```
Kegsay: At the moment you're only really testing that stuff *you* sign will pass your verification process. It isn't testing that the signed JSON is anything other people can understand. Leo made some examples in [the appendices of the spec](https://matrix.org/docs/spec/appendices.html) - It would be nice if you could write a test which used that data and referenced it.

signing_test.go
Kegsay: Looking at his examples, it looks like your sign/verify process will fail for redactable content (which *isn't* signed?).

signing.go
```diff
@@ -0,0 +1,105 @@
...
+	var object map[string]*json.RawMessage
+	var signatures map[string]map[string]Base64String
+	if err := json.Unmarshal(message, &object); err != nil {
```
NegativeMjark: Yes we are mostly signing ``structs``. But turning a struct into ``map[string]*json.RawMessage`` is a faff, and leaks implementation details of the JSON signing into the rest of the code.

```
bytes1, err := json.Marshal(&mystruct.field1)
bytes2, err := json.Marshal(&mystruct.field2)
signed, err := SignJSON(...,map[string]*json.RawMessage{"field1": bytes1, "field2": bytes2})
```
vs
```
bytes, err := json.Marshal(&mystruct)
signed, err := SignJSON(...,bytes)
```

signing.go
NegativeMjark: Fixed 681967c

NegativeMjark: Note that it is ``key_id`` in the spec.

signing_test.go
NegativeMjark: Fixed 	d97e4b6

signing_test.go
NegativeMjark: Done 0e9a965

NegativeMjark:
I think I'm done. @Kegsay PTAL.

signing.go
Kegsay: Okay, I don't really care either way - if you think this form is more helpful to the caller I'm happy with that :smile:

Kegsay:
LGTM!

PR #5, opened 2016-10-19
Functions for fetching keys from a server 

NegativeMjark:
and for checking that the response is valid

### DIFF
--- a/keys.go
+++ b/keys.go
@@ -79,10 +79,10 @@ type KeyChecks struct {
 	MatchingServerName        bool  // Does the server name match what was requested.
 	FutureValidUntilTS        bool  // The valid until TS is in the future.
 	HasEd25519VerifyKey       bool  // The key response has a ED25519 key for the server.
-	ValidEd25519VerifyKey     bool  // The verify keys are valid ED25519 keys.
-	MatchingEd25519Signature  bool  // Every verify key claimed has a valid signature.
+	ValidEd25519VerifyKey     *bool // The verify keys are valid ED25519 keys.
+	MatchingEd25519Signature  *bool // Every verify key claimed has a valid signature.
 	HasTLSFingerprint         bool  // The response includes a TLS fingerprint.
-	ValidSHA256TLSFingerprint bool  // Every TLS fingerprint includes a SHA-256 hash.
+	ValidSHA256TLSFingerprint *bool // Every TLS fingerprint includes a SHA-256 hash.
 	MatchingTLSFingerprint    *bool // The TLS fingerprint for the connection matches one of the listed fingerprints.
 }
 
@@ -101,10 +101,10 @@ func CheckKeys(serverName string, timeNowMs uint64, keys ServerKeys, connState *
 	checks.AllChecksOK = (checks.MatchingServerName &&
 		checks.FutureValidUntilTS &&
 		checks.HasEd25519VerifyKey &&
-		checks.ValidEd25519VerifyKey &&
-		checks.MatchingEd25519Signature &&
+		*checks.ValidEd25519VerifyKey &&
+		*checks.MatchingEd25519Signature &&
 		checks.HasTLSFingerprint &&
-		checks.ValidSHA256TLSFingerprint)
+		*checks.ValidSHA256TLSFingerprint)
 
 	// Only check the fingerprint if we have the TLS connection state.
 	if connState != nil {
@@ -136,22 +136,24 @@ func checkFingerprint(connState *tls.ConnectionState, sha256Fingerprints []Base6
 }
 
 func checkVerifyKeys(keys ServerKeys, checks *KeyChecks) map[string]Base64String {
-	checks.ValidEd25519VerifyKey = true
-	checks.MatchingEd25519Signature = true
+	validEd25519VerifyKey := true
+	matchingEd25519Signature := true
 	verifyKeys := map[string]Base64String{}
 	for keyID, keyData := range keys.VerifyKeys {
 		algorithm := strings.SplitN(keyID, ":", 2)[0]
 		publicKey := keyData.Key
 		if algorithm == "ed25519" {
 			if len(publicKey) != 32 {
-				checks.ValidEd25519VerifyKey = false
+				validEd25519VerifyKey = false
 				continue
 			}
 			if err := VerifyJSON(keys.ServerName, keyID, []byte(publicKey), keys.Raw); err != nil {
-				checks.MatchingEd25519Signature = false
+				matchingEd25519Signature = false
 				continue
 			}
 			checks.HasEd25519VerifyKey = true
+			checks.ValidEd25519VerifyKey = &validEd25519VerifyKey
+			checks.MatchingEd25519Signature = &matchingEd25519Signature
 			verifyKeys[keyID] = publicKey
 		}
 	}
@@ -160,13 +162,14 @@ func checkVerifyKeys(keys ServerKeys, checks *KeyChecks) map[string]Base64String
 
 func checkTLSFingerprints(keys ServerKeys, checks *KeyChecks) []Base64String {
 	var fingerprints []Base64String
-	checks.ValidSHA256TLSFingerprint = true
+	validSHA256TLSFingerprint := true
 	for _, fingerprint := range keys.TLSFingerprints {
 		if len(fingerprint.SHA256) != sha256.Size {
-			checks.ValidSHA256TLSFingerprint = false
+			validSHA256TLSFingerprint = false
 			continue
 		}
 		checks.HasTLSFingerprint = true
+		checks.ValidSHA256TLSFingerprint = &validSHA256TLSFingerprint
 		fingerprints = append(fingerprints, fingerprint.SHA256)
 	}
 	return fingerprints
--- a/keys.go
+++ b/keys.go
@@ -41,8 +41,7 @@ func FetchKeysDirect(serverName, addr, sni string) (*ServerKeys, *tls.Connection
 	}
 	defer tcpconn.Close()
 	tlsconn := tls.Client(tcpconn, &tls.Config{ServerName: sni})
-	err = tlsconn.Handshake()
-	if err != nil {
+	if err = tlsconn.Handshake(); err != nil {
 		return nil, nil, err
 	}
 	connectionState := tlsconn.ConnectionState()
@@ -63,6 +62,7 @@ func FetchKeysDirect(serverName, addr, sni string) (*ServerKeys, *tls.Connection
 	if err != nil {
 		return nil, nil, err
 	}
+	defer response.Body.Close()
 	var keys ServerKeys
 	if keys.Raw, err = ioutil.ReadAll(response.Body); err != nil {
 		return nil, nil, err
--- a/keys.go
+++ b/keys.go
@@ -59,10 +59,12 @@ func FetchKeysDirect(serverName, addr, sni string) (*ServerKeys, *tls.Connection
 
 	// Read the 200 OK from the server.
 	response, err := http.ReadResponse(bufio.NewReader(tlsconn), request)
+	if response != nil {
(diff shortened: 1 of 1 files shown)


keys.go
```diff
@@ -0,0 +1,174 @@
...
+	ValidUntilTS  uint64              `json:"valid_until_ts"` // When this result is valid until in milliseconds.
+	OldVerifyKeys map[string]struct { // Old keys that are now only valid for checking historic events.
+		Key       Base64String `json:"key"`        // The public key.
+		ExpiredTS uint64       `json:"expired_ts"` // When this key stopped being valid for event signing.
+	} `json:"old_verify_keys"`
+}
+
+// FetchKeysDirect fetches the matrix keys directly from the given address.
+// Optionally sets a SNI header if ``sni`` is not empty.
+// Returns the server keys and the state of the TLS connection used to retrive them.
```
Kegsay: `retrive`

keys.go
```diff
@@ -0,0 +1,174 @@
...
+		return nil, nil, err
+	}
+	request.Header.Set("Connection", "close")
+	if err = request.Write(tlsconn); err != nil {
+		return nil, nil, err
+	}
+
+	// Read the 200 OK from the server.
+	response, err := http.ReadResponse(bufio.NewReader(tlsconn), request)
+	if err != nil {
```
Kegsay: Missing `defer response.Close()`. If this is intentional (because of interesting side effects on `connectionState`?) then please comment why you're not closing this response.

keys.go
```diff
@@ -0,0 +1,174 @@
...
+	"io/ioutil"
+	"net"
+	"net/http"
+	"strings"
+)
+
+// ServerKeys are the ed25519 signing keys published by a matrix server.
+// Contains SHA256 fingerprints of the TLS X509 certificates used by the server.
+type ServerKeys struct {
+	Raw             []byte     `json:-`             // Copy of the raw JSON for signature checking.
```
Kegsay: Does that actually work? I thought it was `json:"-"` (with `"`)?

keys.go
```diff
@@ -0,0 +1,174 @@
...
+	return &keys, &connectionState, nil
+}
+
+// KeyChecks are the checks that should be applied to ServerKey responses.
+type KeyChecks struct {
+	AllChecksOK               bool // Did all the checks pass?
+	MatchingServerName        bool // Does the server name match what was requested.
+	FutureValidUntilTS        bool // The valid until TS is in the future.
+	HasEd25519VerifyKey       bool // The key response has a E25519 key for the server.
+	ValidEd25519VerifyKey     bool // The verify keys are value.
```
Kegsay: "are value" :)

keys.go
```diff
@@ -0,0 +1,174 @@
...
+	MatchingEd25519Signature  bool // Every verify key claimed has a valid signature.
+	HasTLSFingerprint         bool // The response includes a TLS fingerprint.
+	ValidSHA256TLSFingerprint bool // Every TLS fingerprint includes a SHA-256 hash.
+	MatchingTLSFingerprint    bool // The TLS fingerprint for the connection matches one of the listed fingerprints.
+}
+
+// CheckKeys checks the keys returned from a server to make sure they are valid.
+// If the checks pass the also returns a map of key_id to Ed25519 public key and a list of SHA256 TLS fingerprints.
```
Kegsay: `If the checks pass the also`?

keys.go
```diff
@@ -0,0 +1,174 @@
...
+	checks.MatchingServerName = serverName == keys.ServerName
+
+	ed25519Keys = checkVerifyKeys(keys, &checks)
+	sha256Fingerprints = checkTLSFingerprints(keys, &checks)
+
+	checks.FutureValidUntilTS = timeNowMs < keys.ValidUntilTS
+
+	if connState == nil {
+		// No TLS connection state so skip the fingerprint checks.
+		checks.MatchingTLSFingerprint = true
```
Kegsay: Is this safe to do? You haven't checked any fingerprints at all, yet the caller doesn't know that: it's being told it matches!

keys.go
```diff
@@ -0,0 +1,174 @@
...
+	} else {
+		// Check the peer certificates.
+		checks.MatchingTLSFingerprint = checkFingerprint(connState, sha256Fingerprints)
```
Kegsay: Given `checkFingerprint` only uses `connState.PeerCertificates`, it feels nicer to just pass in the list of peer certificates (default initialised to 0 items), but YMMV.

keys.go
```diff
@@ -0,0 +1,174 @@
...
+		}
+	}
+	return false
+}
+
+func checkVerifyKeys(keys ServerKeys, checks *KeyChecks) map[string]Base64String {
+	checks.ValidEd25519VerifyKey = true
+	checks.MatchingEd25519Signature = true
+	verifyKeys := map[string]Base64String{}
+	for keyID, keyData := range keys.VerifyKeys {
```
Kegsay: I really dislike not failing safely. If `len(keys.VerifyKeys) == 0` then both `ValidEd25519VerifyKey` and `MatchingEd25519Signature` will be true.

keys.go
```diff
@@ -0,0 +1,174 @@
...
+		algorithm := strings.SplitN(keyID, ":", 2)[0]
+		publicKey := keyData.Key
+		if algorithm == "ed25519" {
```
Kegsay: If this is `false`, both `ValidEd25519VerifyKey` and `MatchingEd25519Signature` will be `true`. The null set shouldn't be truthy :(

keys.go
```diff
@@ -0,0 +1,174 @@
...
+			if len(publicKey) != 32 {
+				checks.ValidEd25519VerifyKey = false
+				continue
+			}
+			if err := VerifyJSON(keys.ServerName, keyID, []byte(publicKey), keys.Raw); err != nil {
+				checks.MatchingEd25519Signature = false
+				continue
+			}
+			checks.HasEd25519VerifyKey = true
```
Kegsay: WRRRRRY! This API sucks and just adds confusion :/ You need to check this field before `ValidEd25519VerifyKey` and `MatchingEd25519Signature` make any sense.

keys.go
```diff
@@ -0,0 +1,174 @@
...
+	}
+	return verifyKeys
+}
+
+func checkTLSFingerprints(keys ServerKeys, checks *KeyChecks) []Base64String {
+	var fingerprints []Base64String
+	checks.ValidSHA256TLSFingerprint = true
+	for _, fingerprint := range keys.TLSFingerprints {
+		if len(fingerprint.SHA256) != sha256.Size {
+			checks.ValidSHA256TLSFingerprint = false
```
Kegsay: Seems like a bit of a weak check. Isn't there something better we can do aside from a basic length check?

keys.go
NegativeMjark: Done d20f4a2dde86999fbcf08a71870e38b27681c65e

keys.go
NegativeMjark: Done d20f4a2dde86999fbcf08a71870e38b27681c65e

keys.go
```diff
@@ -0,0 +1,174 @@
...
+func FetchKeysDirect(serverName, addr, sni string) (*ServerKeys, *tls.ConnectionState, error) {
+	// Create a TLS connection.
+	tcpconn, err := net.Dial("tcp", addr)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer tcpconn.Close()
+	tlsconn := tls.Client(tcpconn, &tls.Config{ServerName: sni})
+	err = tlsconn.Handshake()
+	if err != nil {
```
NegativeMjark: Done d20f4a2dde86999fbcf08a71870e38b27681c65e

keys.go
NegativeMjark: Fixed 9d8bccfb5ffe51a71519dc25bdf6d4fcdf13325d

keys.go
NegativeMjark: Fixed 9c8f161bab2d62c5f81922ea1249524f2f6d3059

keys.go
NegativeMjark: Fixed 9c8f161bab2d62c5f81922ea1249524f2f6d3059

Kegsay:
I have reservations about the API presented by the `KeyChecks` struct. I don't like the form `HasXXX` then `XXXIsValid` because the latter only makes sense if the former is `true`. This is done in two places:
 - The `Ed25519VerifyKey`
 - The `SHA256TLSFingerprint`

keys.go
NegativeMjark: I'd rather not use a nil/empty slice for no certificates here since someone might unpack a zero value ConnectionState and end up passing an empty list by mistake.

NegativeMjark: Also I might want to extend the checks to include checking the connection CipherSuite at somepoint.

keys.go
NegativeMjark: Fixed a742441

keys.go
NegativeMjark: Fixed a742441

keys.go
NegativeMjark: Fixed a742441

keys.go
NegativeMjark: Nope.

keys.go
```diff
@@ -0,0 +1,176 @@
...
+	ed25519Keys = checkVerifyKeys(keys, &checks)
+	sha256Fingerprints = checkTLSFingerprints(keys, &checks)
+
+	checks.FutureValidUntilTS = timeNowMs < keys.ValidUntilTS
+
+	checks.AllChecksOK = (checks.MatchingServerName &&
+		checks.FutureValidUntilTS &&
+		checks.HasEd25519VerifyKey &&
+		*checks.ValidEd25519VerifyKey &&
+		*checks.MatchingEd25519Signature &&
```
Kegsay: NPE: This will panic if the checks are not done.

keys.go
```diff
@@ -0,0 +1,176 @@
...
+	validSHA256TLSFingerprint := true
+	for _, fingerprint := range keys.TLSFingerprints {
+		if len(fingerprint.SHA256) != sha256.Size {
+			validSHA256TLSFingerprint = false
+			continue
+		}
+		checks.HasTLSFingerprint = true
+		checks.ValidSHA256TLSFingerprint = &validSHA256TLSFingerprint
+		fingerprints = append(fingerprints, fingerprint.SHA256)
+	}
```
Kegsay:
```go
// emph mine
ValidSHA256TLSFingerprint *bool // **Every** TLS fingerprint includes a SHA-256 hash.
```

If `keys.TLSFingerprints` has 2 entries: `[good, bad]` then:
 - The first loop succeeds and sets the `ValidSHA256TLSFingerprint` to true
 - The second loop fails the length check and `continue`s, there's no more entries so the for loop exits and returns `ValidSHA256TLSFingerprint` as `true`.

NegativeMjark:
```go
checks.ValidSHA256TLSFingerprint = &validSHA256TLSFingerprint
```
That doesn't set the ValidSHA256TLSFingerprint to true.
It sets it to a pointer to validSHA256TLSFingerprint which will then be set to false.

keys.go
```diff
@@ -0,0 +1,176 @@
...
+	if err = request.Write(tlsconn); err != nil {
+		return nil, nil, err
+	}
+
+	// Read the 200 OK from the server.
+	response, err := http.ReadResponse(bufio.NewReader(tlsconn), request)
+	if err != nil {
+		return nil, nil, err
+	}
+	defer response.Body.Close()
```
NegativeMjark: Fixed 2cd486f

keys.go
Kegsay: Aha, yes that makes sense. This still fails if `keys.TLSFingerprints` has 1 entry: [bad] though.

Kegsay:
LGTM - Tests can come at a later date :)

PR #8, opened 2016-11-29
Add Client type

Kegsay:
Manually tested with my open openid token and it works on matrix.org.

### DIFF
--- a/client.go
+++ b/client.go
@@ -38,7 +38,9 @@ func NewClient() *Client {
 				}
 				// Wrap a raw connection ourselves since tls.Dial defaults the SNI
 				conn := tls.Client(rawconn, &tls.Config{
-					ServerName:         "",
+					ServerName: "",
+					// TODO: We should be checking that the TLS certificate we see here matches
+					//       one of the allowed SHA-256 fingerprints for the server.
 					InsecureSkipVerify: true,
 				})
 				if err := conn.Handshake(); err != nil {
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -0,0 +1,127 @@
...
+	err = fmt.Errorf("no address found for matrix host %v", host)
+	for _, addr := range dnsResult.Addrs {
+		u := makeHTTPSURL(r.URL, addr)
+		r.URL = &u
+		resp, err = f.transport.RoundTrip(r)
+		if err == nil {
+			return resp, nil
+		}
+	}
+	return nil, err
```
NegativeMjark: shouldn't the ``err = fmt.Errorf`` be here rather than before the for loop?

client.go
```diff
@@ -0,0 +1,127 @@
...
+			// as the empty string so that crypto/tls doesn't add SNI.
+			DialTLS: func(network, addr string) (net.Conn, error) {
+				rawconn, err := net.Dial(network, addr)
+				if err != nil {
+					return nil, err
+				}
+				// Wrap a raw connection ourselves since tls.Dial defaults the SNI
+				conn := tls.Client(rawconn, &tls.Config{
+					ServerName:         "",
+					InsecureSkipVerify: true,
```
NegativeMjark: We should be checking that the TLS certificate we see here matches one of the allowed SHA-256 fingerprints for the server.

That might be a bit much for the initial implementation, but a ``TODO`` comment might be nice.

Kegsay: Done.

client.go
Kegsay: Done.

PR #10, opened 2017-01-06
Add golint and go vet to pre-commit hook

NegativeMjark:


hooks/pre-commit
```diff
@@ -2,8 +2,9 @@
...
 set -eu
 
-golint
+golint ./...
 go fmt
+go tool vet .
 go tool vet --shadow .
-gocyclo -over 12 .
-go test
+gocyclo -over 19 .
```
Kegsay: Why?

hooks/pre-commit
```diff
@@ -2,8 +2,9 @@
 
 set -eu
 
-golint
+golint ./...
 go fmt
+go tool vet .
 go tool vet --shadow .
```
Kegsay: uhhh... `go tool vet` is already being called?

hooks/pre-commit
```diff
@@ -2,8 +2,9 @@
...
-gocyclo -over 12 .
-go test
+gocyclo -over 19 .
+go test .
```
Kegsay: The `.` shouldn't be necessary (works fine without on `gomatrix` which uses the same directory structure).

Also, you probably want to make this:
```
go test -timeout 5s -test.v
```

which sets a 5s timeout for each test and logs test names and stdout which can be useful when debugging problems

hooks/pre-commit
NegativeMjark: --shadow seems to turn of the other checks?

Kegsay:
```
  -shadow
    	check for shadowed variables (experimental; must be set explicitly)
```

Kegsay: Oh fun.
```
By default, -all is set and all non-experimental checks are run.
For more information run
	go doc cmd/vet

Flags:
  -all
    	enable all non-experimental checks
```

So setting `-shadow` turns OFF `-all` :/

Kegsay:
LGTM

PR #12, opened 2017-01-09
Implement the checks for m.room.create events.

NegativeMjark:


eventauth/eventauth.go
```diff
@@ -199,8 +200,39 @@ func Allowed(event Event, authEvents AuthEvents) error {
...
+	}
+	senderDomain, err := domainFromID(event.Sender)
+	if err != nil {
+		return err
+	}
+	if senderDomain != roomIDDomain {
+		return errorf("create event room ID domain does not match sender: %q != %q", roomIDDomain, senderDomain)
+	}
+	if len(event.PrevEvents) > 0 {
+		return errorf("create event must be the first event in the room")
```
Kegsay: It would be polite to state how many previous events there were.

eventauth/eventauth.go
```diff
@@ -199,8 +200,39 @@ func Allowed(event Event, authEvents AuthEvents) error {
 	}
 }
 
+// createEventAllowed checks whether the m.room.create event is allowed.
+// It returns an error if the event is not allowed.
 func createEventAllowed(event Event, authEvents AuthEvents) error {
```
Kegsay: We don't use `authEvents`. Why is it being passed as a parameter?

eventauth/eventauth_test.go
```diff
@@ -116,3 +116,132 @@ func TestStateNeededForInvite3PID(t *testing.T) {
...
+	return &event, nil
+}
+
+type testCase struct {
+	AuthEvents testAuthEvents    `json:"auth_events"`
+	Allowed    []json.RawMessage `json:"allowed"`
+	NotAllowed []json.RawMessage `json:"not_allowed"`
+}
+
+func testEventAuth(t *testing.T, testCaseData string) {
```
Kegsay: You mean `testCaseJSON`?

Kegsay:
Other than minor nits, LGTM!

PR #13, opened 2017-01-09
Implement the default checks for normal events.

NegativeMjark:


eventauth/eventauth.go
```diff
@@ -109,12 +108,6 @@ func StateNeededForAuth(events []Event) (result StateNeeded) {
 	return
 }
 
-// memberContent is the JSON content of a m.room.member event.
-type memberContent struct {
-	Membership       string          `json:"membership"`
-	ThirdPartyInvite json.RawMessage `json:"third_party_invite"`
-}
```
NegativeMjark: Moved to eventcontent.go

eventauth/eventauth.go
```diff
@@ -220,21 +213,6 @@ func createEventAllowed(event Event) error {
...
-	// Split on the first ":" character since the domain can contain ":"
-	// characters.
-	parts := strings.SplitN(id, ":", 2)
-	if len(parts) != 2 {
-		// The ID must have a ":" character.
-		return "", errorf("invalid ID: %q", id)
-	}
-	// Return everything after the first ":" character.
-	return parts[1], nil
-}
```
NegativeMjark: Moved to eventcontent.go

eventauth/eventcontent.go
```diff
@@ -0,0 +1,280 @@
+package eventauth
+
+import (
+	"encoding/json"
+	"strconv"
+	"strings"
+)
+
+// creaateContent is the JSON content of a m.room.create event along with
```
Kegsay: creaateContent

eventauth/eventcontent.go
```diff
@@ -0,0 +1,280 @@
...
+// the top level keys needed for auth.
+type createContent struct {
+	// We need the domain of the create event when checking federatebility.
+	senderDomain string
+	// We need the roomID to check that events are in the same room as the create event.
+	roomID string
```
Kegsay: Why is this here? I am grateful for the explanation so I know what it is used for, but it seems weird to have this here. Any justification?

eventauth/eventcontent.go
```diff
@@ -0,0 +1,280 @@
...
+	// The "m.federate" flag tells us whether the room can be federated to other servers.
+	Federate *bool `json:"m.federate"`
+	// The creator of the room tells us what the default power levels are.
```
Kegsay: Really? I thought the `creator` was the ID of the user who called `/createRoom`.

eventauth/eventcontent.go
```diff
@@ -0,0 +1,280 @@
...
+func (c *createContent) domainAllowed(domain string) error {
+	if domain == c.senderDomain {
+		// If the domain matches the domain of the create event
+		return nil
+	}
+	if c.Federate == nil || *c.Federate {
+		// The m.federate field defaults to true.
+		return nil
+	}
+	return errorf("room is unfederatable")
```
Kegsay: Yucky, I don't like the style of "filter out successes then what's left is an error" because it is really error-prone when compared to "filter out errors then what's left is a success". For example, what happens in this function if `domain != c.senderDomain` and `c.Federate = true`? The domain will be **allowed** even though the sender domain doesn't match!

eventauth/eventcontent.go
```diff
@@ -0,0 +1,280 @@
...
+}
+
+// idAllowed checks whether the domain of the ID is allowed in the room.
```
Kegsay: I believe you're being purposefully vague because this will work for all of the Matrix-style IDs, but it doesn't help people understand what they can or cannot reasonably pass into this function. I would prefer it if this explicitly stated room and user IDs, given this will totally work for room aliases, event IDs, etc which is presumably not what your intention is here.

eventauth/eventcontent.go
NegativeMjark: This style is superior cause it list the way the event is allowed. Rather than trying to list all the ways the event is *not* allowed which is liable to miss cases.

But I take your point this bit needs better comments.

eventauth/eventcontent.go
```diff
@@ -0,0 +1,280 @@
...
+	parts := strings.SplitN(id, ":", 2)
+	if len(parts) != 2 {
+		// The ID must have a ":" character.
+		return "", errorf("invalid ID: %q", id)
+	}
+	// Return everything after the first ":" character.
+	return parts[1], nil
+}
+
+// memberContent is the JSON content of a m.room.member event.
```
Kegsay: Again, I'm struggling to conceptualise what `memberContent` is actually supposed to be representing. It sure isn't what you're stating here, because it's missing a bunch of fields from http://matrix.org/docs/spec/client_server/r0.2.0.html#m-room-member - given this is for *auth checks* I imagine you only care about the fields which are used for auth, so you should mention this in your description.

eventauth/eventcontent.go
```diff
@@ -0,0 +1,280 @@
...
+type memberContent struct {
+	Membership       string          `json:"membership"`
+	ThirdPartyInvite json.RawMessage `json:"third_party_invite"`
+}
+
+// Load the member content from the member event for the user ID.
+// Returns an error if there was an error loading the member event or
+// parsing the event content.
+func (c *memberContent) load(authEvents AuthEvents, userID string) error {
```
Kegsay: You have a `load` and `parse` function for this struct. That's confusing. You are consistently passing `authEvents` to all functions named `load`, so perhaps you should make this clearer e.g `loadFromAuthEvents` or to be more terse: `fromAuthEvents`.

eventauth/eventcontent.go
```diff
@@ -0,0 +1,280 @@
...
+	memberEvent, err := authEvents.Member(userID)
+	if err != nil {
+		return err
+	}
+	if memberEvent == nil {
+		c.Membership = "leave"
```
Kegsay: This warrants a tiny explanatory comment to explain why.

eventauth/eventcontent.go
```diff
@@ -0,0 +1,280 @@
...
+
+}
+
+// parse the power level content from an event.
+func (c *powerLevelContent) parse(event Event) error {
+	// Set the levels to their default values.
+	c.defaults()
+
+	// We can't extract the JSON directly to the powerLevelContent because we
+	// need to convert string values to int values.
```
Kegsay: This should probably also be mentioned in the struct definition. I nearly whined at you for it.

eventauth/eventcontent.go
NegativeMjark: Hmm, it's interesting. The auth code needs both top-level keys from the create event object, and the parsed create event content.
It turned out easier to write that code by shuffling the top-level keys into the same object as the create event content, rather than passing the two around separately.

eventauth/eventcontent.go
NegativeMjark: That is also true. I was documenting why I was extracting the key rather than repeating the documentation from the spec on where the key comes form. Do you think I linking to the spec would help?

eventauth/eventcontent.go
NegativeMjark: I was being purposely vague. But on closer inspection I'm only going to use this with user IDs so I might as well document it as such. 7c32ab3

eventauth/eventcontent.go
NegativeMjark: Done 7c32ab3

eventauth/eventcontent.go
NegativeMjark: Done d87fb92

eventauth/eventcontent.go
Kegsay: I wouldn't assume prior knowledge when it isn't self-explanatory - `creator` isn't clear if this is a user ID, server domain, or something completely different. A spec link would help clear up ambiguity and provide a reference for explaining what the field is for, so SGTM.

eventauth/eventcontent.go
Kegsay: I thought so, and yeah it's an interesting one. I think it can be confusing to pollute namespaces like this, but I understand that it may be even worse to have to pass around the same 2 structs to everything. Let's leave it for now and see how bad this gets as you add more logic.

eventauth/eventauth_test.go
```diff
@@ -203,7 +203,7 @@ func testCaseJSON(t *testing.T, testCaseData string) {
 	for _, data := range tc.NotAllowed {
 		var event Event
 		if err := json.Unmarshal(data, &event); err != nil {
-			t.Fatal(err)
+			panic(err)
```
Kegsay: Any reasons for this? I thought `.Fatal` did panic?

eventauth/eventauth.go
```diff
@@ -251,6 +229,73 @@ func redactEventAllowed(event Event, authEvents AuthEvents) error {
 	panic("Not implemented")
 }
 
+// defaultEventAllowed checks whether the event is allowed by the default
+// checks for events.
+// It returns an error if the event is not allowed or if there was a
+// problem loading the auth events needed.
```
Kegsay: Do we really want to bundle explicit event rejection into the same error response as "general problems loading the auth events"? I worry the latter may be transient, but the former isn't?

eventauth/eventauth.go
```diff
@@ -251,6 +229,73 @@ func redactEventAllowed(event Event, authEvents AuthEvents) error {
...
+// An eventAllower has the information needed to authorise all events types
+// other than m.room.create, m.room.member and m.room.alias which are special.
+type eventAllower struct {
+	create      createContent
+	member      memberContent
+	powerLevels powerLevelContent
+}
+
+// setup loads the necesary information from the auth events.
+func (e *eventAllower) setup(authEvents AuthEvents, senderID string) error {
```
Kegsay: > necesary

Also, in general I thought "in order to set up this struct you need to execute this arbitrary code" stuff is typically done via a `New()` call which sets up the struct and returns the struct possibly with an error. Any reason for deviating from that convention here? Is it ever valid to be using an `eventAllower` *without* calling `setup()`?

eventauth/eventauth.go
```diff
@@ -251,6 +229,73 @@ func redactEventAllowed(event Event, authEvents AuthEvents) error {
...
+	if err := e.create.load(authEvents); err != nil {
+		return err
+	}
+	if err := e.member.load(authEvents, senderID); err != nil {
+		return err
+	}
+	if err := e.powerLevels.load(authEvents, e.create.Creator); err != nil {
```
Kegsay: In a similar vein to https://github.com/matrix-org/gomatrixserverlib/pull/13/files#r95351964 - should we not have `New()` calls for these things as well?

eventauth/eventauth.go
```diff
@@ -251,6 +229,73 @@ func redactEventAllowed(event Event, authEvents AuthEvents) error {
...
+		return err
+	}
+	return nil
+}
+
+// commonChecks does the checks that are applied to all events types other than
```
Kegsay: > events

eventauth/eventauth.go
```diff
@@ -251,6 +229,73 @@ func redactEventAllowed(event Event, authEvents AuthEvents) error {
...
+	if event.RoomID != e.create.roomID {
+		return errorf("create event has different roomID: %q != %q", event.RoomID, e.create.roomID)
+	}
+
+	if err := e.create.userIDAllowed(event.Sender); err != nil {
+		return err
+	}
+
+	if e.member.Membership != "join" {
+		return errorf("sender %q not in room", event.Sender)
```
Kegsay: How does this work on transition states? E.g. invite -> join and join -> leave? Comment please.

eventauth/eventcontent.go
```diff
@@ -0,0 +1,290 @@
...
+		return level
+	}
+	return c.userDefaultLevel
+}
+
+// eventLevel returns the power level needed to send an event in the room.
+func (c *powerLevelContent) eventLevel(eventType string, eventStateKey *string) int64 {
+	if eventType == "m.room.third_party_invite" {
+		// Special case third_party_invite events to have the same level as
+		// m.room.member invite events.
```
Kegsay: As per [the spec](http://matrix.org/docs/spec/client_server/r0.2.0.html#m-room-third-party-invite) - this event is a state event but you do not assert this is the case.

eventauth/eventcontent.go
```diff
@@ -0,0 +1,290 @@
...
+	powerLevelsEvent, err := authEvents.PowerLevels()
+	if err != nil {
+		return err
+	}
+	if powerLevelsEvent != nil {
+		return c.parse(*powerLevelsEvent)
+	}
+
+	// If there are no power leves then fall back to defaults.
+	c.defaults()
```
Kegsay: Comment suggests there should be a conditional but there isn't one. Also:

> leves

eventauth/eventcontent_test.go
```diff
@@ -0,0 +1,36 @@
...
+		if int64(want) != got.value {
+			t.Fatalf("Wanted %d got %q", want, got.value)
+		}
+	}
+}
+
+func TestLevelJSONValueInvalid(t *testing.T) {
+	var values []levelJSONValue
+	inputs := []string{
+		`[{}]`, `[[]]`, `["not a number"]`, `["0.0"]`,
```
Kegsay: Oh, surprising! Why does 0.0 fail? Shouldn't that be 0? Is this for parity with python's `int()` impl?

eventauth/eventcontent.go
```diff
@@ -0,0 +1,290 @@
...
+	if len(parts) != 2 {
+		// The ID must have a ":" character.
+		return "", errorf("invalid ID: %q", id)
+	}
+	// Return everything after the first ":" character.
+	return parts[1], nil
+}
+
+// memberContent is the JSON content of a m.room.member event needed for
+// for auth checks.
```
Kegsay: > needed for for auth

eventauth/eventauth_test.go
```diff
@@ -234,14 +236,282 @@ func TestAllowedEmptyRoom(t *testing.T) {
...
+		"allowed": [{
+			"type": "m.room.message",
+			"sender": "@u1:a",
+			"room_id": "!r1:a",
+			"event_id": "$e3:a",
+			"content": {"body": "Test"}
+		}],
+		"not_allowed": [{
+			"type": "m.room.message",
+			"sender": "@u2:b",
```
Kegsay: I'd prefer if you actually just varied the server rather than the local part as well here, so `@u1:b` because it's the domain part we care about.

eventauth/eventauth_test.go
```diff
@@ -234,14 +236,282 @@ func TestAllowedEmptyRoom(t *testing.T) {
...
+			}
+		}, {
+			"type": "my.custom.state",
+			"state_key": "@u2:a",
+			"sender": "@u1:a",
+			"room_id": "!r1:a",
+			"event_id": "@e17:a",
+			"content": {},
+			"unsigned": {
+				"not_allowed": "State key starts with '@' and is for a different user"
```
Kegsay: :1st_place_medal: for consistently doing this on the tests, this makes them so nice to read ❤️

eventauth/eventcontent_test.go
NegativeMjark: "0.0" should fail because ```int("0.0")``` is not the same as ```int(float("0.0"))``` in python.

eventauth/eventauth_test.go
NegativeMjark: panic gives me more interesting backtraces.

Kegsay: Fair enough!

eventauth/eventcontent_test.go
Kegsay: Thought so, okay then.

eventauth/eventcontent.go
Kegsay: Hmm, okay, maybe I just got confused then.

NegativeMjark:
I think it would be better to hold off implementing the state_key checks for now since synapse doesn't perform any and I'd prefer that we kept the divergence from synapse's auth checks to a minimum.

NegativeMjark:
Otherwise @Kegsay PTAL

Kegsay:
> I think it would be better to hold off implementing the state_key checks for now since synapse doesn't perform any and I'd prefer that we kept the divergence from synapse's auth checks to a minimum.

That's fine, provided you put in enormous shouty caps that this is a known thing we're not addressing and the explanation as to why. Basically, what you've just said, but in comment form.

Kegsay:
Other than that, LGTM.

NegativeMjark:
Comment is added. @Kegsay are you happy?

Kegsay:
LGTM

PR #15, opened 2017-01-13
Implement the auth checks for m.room.member events

NegativeMjark:


eventauth/eventauth.go
```diff
@@ -219,8 +219,14 @@ func createEventAllowed(event Event) error {
 	return nil
 }
 
+// memberAllowed checks whether the m.room.member event is allowed.
```
Kegsay: I think you mean `memberEventAllowed`

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+}
+
+// newMembershipAllower loads the information needed to authenticate the m.room.member event
+// from the auth events.
+func newMembershipAllower(authEvents AuthEvents, event Event) (m membershipAllower, err error) {
+	if event.StateKey == nil {
+		err = errorf("m.room.member must be a state event")
+		return
+	}
+	m.targetID = *event.StateKey
```
Kegsay: It's worrying that we do no assertions to make sure the state key string looks like a user ID.

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+		return
+	}
+	if m.senderMember, err = newMemberContentFromAuthEvents(authEvents, m.senderID); err != nil {
+		return
+	}
+	if m.powerLevels, err = newPowerLevelContentFromAuthEvents(authEvents, m.create.Creator); err != nil {
+		return
+	}
+	if m.newMember.Membership == "join" {
+		if m.joinRule, err = newJoinRuleContentFromAuthEvents(authEvents); err != nil {
```
Kegsay: Huh? We aren't doing join_rule checks here, so why are we only loading this if `newMember.Membership == "join"`?

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
 
 	return nil
 }
+
+// A membershipAllower has the information needed to authenticate a m.room.member event
+type membershipAllower struct {
+	// The user ID of the user who's membership is changing.
```
ara4n: whose

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+	targetID string
+	// The user ID of the user who sent the membership event.
+	senderID string
+	// The membership of the user who sent the membership event.
+	senderMember memberContent
+	// The previous membership of the user who's membership is changing.
```
ara4n: whose

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+	}
+	// Special case the first join event in the room to allow the creator to join.
+	// https://github.com/matrix-org/synapse/blob/v0.18.5/synapse/api/auth.py#L328
+	if m.targetID == m.create.Creator &&
+		m.newMember.Membership == "join" &&
+		m.senderID == m.targetID &&
+		len(event.PrevEvents) == 1 {
+		// prev_events is a list of 2 element lists of event IDs and hashes.
+		if len(event.PrevEvents[0]) != 2 {
+			return errorf("unparsable prev event")
```
Kegsay: Why on earth does the `membershipAllower` need to know how to parse `PrevEvents`? I'd MUCH rather if there is an `event.PrevEvents` that it was validated/parsed when it was assigned to `event`, not randomly here.

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+		}
+		// Unmarshall the event ID string.
```
Kegsay: > Unmarshall

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+	// Otherwise this is an attempt to modify the membership of somebody else.
+	return m.membershipAllowedOther()
+}
+
+// membershipAllowedSelf determines if the change made by the user to their own membership is allowed.
+func (m *membershipAllower) membershipAllowedSelf() error {
+	if m.newMember.Membership == "join" {
+		// A user that is not in the room is allowed to join if the room
+		// join rules are "public".
+		if m.oldMember.Membership == "leave" && m.joinRule.JoinRule == "public" {
```
Kegsay: I'd appreciate it if you could yank `"leave"` and co out to be constants so the compiler will whine if we mis-spell.

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+			m.targetID, m.oldMember.Membership, m.newMember.Membership,
+		)
+	}
+
+	if m.senderMember.Membership != "join" {
+		return errorf("sender %q is not in the room", m.senderID)
+	}
+
+	return errorf(
+		"%q is not allowed to change the membership of %q frEventom %q to %q",
```
Kegsay: > frEventom

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+		}
+		// An invited user is allowed to reject an invite.
+		if m.oldMember.Membership == "invite" {
+			return nil
+		}
+	}
+	return m.membershipFailed()
+}
+
+// membershipAllowedSelf determines if the user is allowed to change the membership of another user.
```
Kegsay: > membershipAllowedSelf

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+func (m *membershipAllower) membershipAllowedOther() error {
+	senderLevel := m.powerLevels.userLevel(m.senderID)
+	targetLevel := m.powerLevels.userLevel(m.targetID)
+
+	// You may only modify the membership of another user if you are in the room.
+	if m.senderMember.Membership == "join" {
```
Kegsay: Invert and outdent please:
```go
// You may only modify the membership of another user if you are in the room.
if m.senderMember.Membership != "join" {
    return m.membershipFailed()
}
```

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+		if m.newMember.Membership == "ban" {
+			// A user may ban another user if their level is high enough
+			if senderLevel >= m.powerLevels.banLevel &&
+				senderLevel > targetLevel {
```
Kegsay: Please cite the line in Synapse because it's non-obvious that:

> A user may ban another user if their level is high enough

should have the `senderLevel > targetLevel` check.

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+				return nil
+			}
+			// A user may kick another user if their level is high enough.
+			if m.oldMember.Membership != "ban" &&
+				senderLevel >= m.powerLevels.kickLevel &&
+				senderLevel > targetLevel {
+				return nil
+			}
+		}
+		if m.newMember.Membership == "invite" {
```
Kegsay: Nit: I wonder if a switch for `m.newMember.Membership` would be clearer to read, as it makes it more obvious that only 1 of these if branches can possibly execute. Currently I have to double-check all three if statements to make sure there isn't a sneaky change from `m.newMember.Membership` to `m.oldMember.Membership` or something equally odd, and it's more cognitive burden than I'd like.

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+				return nil
+			}
+		}
+		if m.newMember.Membership == "leave" {
+			// A user may unban another user if their level is high enough.
+			if m.oldMember.Membership == "ban" && senderLevel >= m.powerLevels.banLevel {
+				return nil
+			}
+			// A user may kick another user if their level is high enough.
+			if m.oldMember.Membership != "ban" &&
```
Kegsay: This will fire for cases where A tries to kick B, when B had already left the room (state `leave`). That feels weird, and certainly warrants some prose explaining either way.

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+				senderLevel >= m.powerLevels.kickLevel &&
+				senderLevel > targetLevel {
+				return nil
```
Kegsay: I don't like the inside of the `if` block having the same amount of indentation as the `if` conditional. Can we double-indent, or convince me that this isn't hard to read?

eventauth/eventauth.go
```diff
@@ -362,3 +368,199 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+func (m *membershipAllower) membershipFailed() error {
+	if m.senderID == m.targetID {
+		return errorf(
+			"%q is not allowed to change their membership from %q to %q",
+			m.targetID, m.oldMember.Membership, m.newMember.Membership,
+		)
+	}
+
+	if m.senderMember.Membership != "join" {
+		return errorf("sender %q is not in the room", m.senderID)
```
Kegsay: I feel it would be polite to tell the client what their membership actually *is*.

eventauth/eventauth_test.go
```diff
@@ -255,6 +255,15 @@ func TestAllowedEmptyRoom(t *testing.T) {
 				"not_allowed": "No create event"
 			}
 		}, {
+			"type": "m.room.member",
+			"sender": "@u1:a",
+			"room_id": "!r1:a",
+			"event_id": "$e4:a",
+			"content": {"membership": "join"},
+			"unsigned": {
+				"not_allowed": "No create event"
```
Kegsay: This has no `state_key` so is a bad test example.

NegativeMjark: Fixed a9fde1c

eventauth/eventauth.go
```diff
@@ -362,3 +376,200 @@ func (e *eventAllower) commonChecks(event Event) error {
...
+	if err := m.create.userIDAllowed(m.targetID); err != nil {
+		return err
+	}
+	// Special case the first join event in the room to allow the creator to join.
+	// https://github.com/matrix-org/synapse/blob/v0.18.5/synapse/api/auth.py#L328
+	if m.targetID == m.create.Creator &&
+		m.newMember.Membership == join &&
+		m.senderID == m.targetID &&
+		len(event.PrevEvents) == 1 {
+		// We unmarshall the prev_events here because this is the only place that
```
Kegsay: > unmarshall

NegativeMjark:
@Kegsay I initially tried writing the membership other checks as switch statements. But I found it harder to read because it was harder to verify the checks being applied in each case.

The advantage of the if statements is that you can start from the "nil" case and scan backwards to see what all the conditionals are. The switch statements make that harder to do in my opinion.

Here's what the switch stmt version looks like for comparison:

```
switch m.newMember.Membership {
	case ban:
		// A user may ban another user if their level is high enough
		if senderLevel >= m.powerLevels.banLevel &&
			senderLevel > targetLevel {
			return nil
		}
	case leave:
		switch m.oldMember.Membership {
		case ban:
			// A user may unban another user if their level is high enough.
			if senderLevel >= m.powerLevels.banLevel {
				return nil
			}
		default:
			// A user may kick another user if their level is high enough.
			if senderLevel >= m.powerLevels.kickLevel && senderLevel > targetLevel {
				return nil
			}
		}
	case invite:
		switch m.oldMember.Membership {
		case leave:
			// A user may invite another user if the user has left the room.
			// and their level is high enough.
			if senderLevel >= m.powerLevels.inviteLevel {
				return nil
			}
		case invite:
			// A user may re-invite a user.
			if senderLevel >= m.powerLevels.inviteLevel {
				return nil
			}
		}
	}
```

eventauth/eventauth.go
NegativeMjark: Yeah, Maybe we should start a tracking these omissions somewhere, I'd quite like to update this and synapse's checks in parallel to keep them in sync.

Kegsay: Add `TODO`s?

Kegsay:
As per IRL: I question the value / how often people will backtrack from a "known" `nil` case to work out the conditionals required to satisfy it, but equally I think either format is clear, so I'm happy to leave it as is if you prefer the `if` form.

eventauth/eventauth.go
Kegsay: Apparently this is just what `go fmt` does :/

Kegsay:
LGTM

PR #18, opened 2017-01-20
Implement matrix event signing and redaction

NegativeMjark:


eventcrypto.go
```diff
@@ -0,0 +1,161 @@
...
+import (
+	"bytes"
+	"crypto/sha256"
+	"encoding/json"
+	"fmt"
+	"golang.org/x/crypto/ed25519"
+)
+
+// AddContentHashesToEvent sets the "hashes" key of the event with a SHA-256 hash of the unredacted event content.
+// This hash is used to detect whether the unredacted content of the event is valid.
```
Kegsay: Would be nice to state that you're returning the event with said hashes somewhere, it's all rather implicit atm and you can't use the types to infer it since it's just `[]byte]`.

redactevent.go
```diff
@@ -0,0 +1,131 @@
...
+// a value type
+type rawJSON []byte
+
+// MarshalJSON implements the json.Marshaller interface using a value receiver.
+// This means that rawJSON used as an embedded value will still encode correctly.
+func (r rawJSON) MarshalJSON() ([]byte, error) {
+	return []byte(r), nil
+}
+
+// UnmarshalJSON impelements the json.Unmarshaller interface using a pointer receiver.
```
Kegsay: > impelements

redactevent.go
```diff
@@ -0,0 +1,131 @@
...
+import (
+	"encoding/json"
+)
+
+// rawJSON is a reimplementation of json.RawMessage that supports being used as
+// a value type
+type rawJSON []byte
+
+// MarshalJSON implements the json.Marshaller interface using a value receiver.
+// This means that rawJSON used as an embedded value will still encode correctly.
```
Kegsay: You're going to need to give some usage examples here in order to convince me why you can't just use `json.RawMessage`. I *think* I know why (it incorrectly encodes as base 64) but given you're re-implementing parts of the stdlib, I think it warrants a better, more detailed explanation.

redactevent.go
```diff
@@ -0,0 +1,131 @@
...
+
+	// memberContent keeps the fields needed in a m.room.member event.
+	// Member events keep the membership.
+	// (In an ideal world they would keep the third_party_invite see matrix-org/synapse#1831)
+	type memberContent struct {
+		Membership rawJSON `json:"membership,omitempty"`
+	}
+
+	// aliasesContent keeps the fields needed in a m.room.aliases event.
+	// Alias events probably don't need to keep the aliases key, but we need to match synapse here.
```
Kegsay: Can you TODO marker this please so we can revisit this decision.

redactevent.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+
+	// historyVisibilityContent keeps the fields needed in a m.room.history_visibility event
+	// History visibility events need to keep the history_visibility key.
+	type historyVisibilityContent struct {
+		HistoryVisibility rawJSON `json:"history_visibility,omitempty"`
+	}
+
+	// allContent keeps the union of all the content fields needed across all the event types.
+	//All the content JSON keys we are keeping are distinct across the different event types.
```
Kegsay: Missing space after `//`.

redactevent.go
```diff
@@ -0,0 +1,131 @@
...
+		OriginServerTS rawJSON    `json:"origin_server_ts,omitempty"`
+		Membership     rawJSON    `json:"membership,omitempty"`
+	}
+
+	var event eventFields
+	// Unmarshalling into a struct will discard any extra fields from the event.
+	if err := json.Unmarshal(eventJSON, &event); err != nil {
+		return nil, err
+	}
+	// Create a new, empty content object
```
Kegsay: Unhelpful comment, I can see what you're doing in the code. **Why** are we making a new one when there is a populated one from unmarshalling `eventFields`.

Kegsay: In fact, `:110-:111` explain this good enough, this comment adds nothing IMO.

redactevent.go
```diff
@@ -0,0 +1,131 @@
...
+	var newContent allContent
+	// Copy the content fields that we should keep for the event type.
+	// By default we copy nothing leaving the content object empty.
+	switch event.Type {
+	case "m.room.create":
+		newContent.createContent = event.Content.createContent
+	case "m.room.member":
```
Kegsay: I think I mentioned this before, but I would really like to see these yanked out into constants so we can't fat-finger the event type strings...

redactevent.go
```diff
@@ -0,0 +1,131 @@
...
+		memberContent
+		aliasesContent
+		historyVisibilityContent
+	}
+
+	// eventFields keeps the top level keys needed by all event types.
+	// (In an ideal world they would include the "redacts" key for m.room.redaction events, see matrix-org/synapse#1831)
+	type eventFields struct {
+		EventID        rawJSON    `json:"event_id,omitempty"`
+		Sender         rawJSON    `json:"sender,omitempty"`
```
Kegsay: http://matrix.org/docs/spec/client_server/r0.2.0.html#redactions says `user_id` not `sender`. Please cite a Synapse line to prove otherwise.

eventcrypto.go
Kegsay: Also, this comment is misleading. AFAICT this has nothing to do with `event.content`, it's adding hashes for the entire event. Am I wrong?

eventcrypto.go
```diff
@@ -0,0 +1,161 @@
...
+	var event map[string]rawJSON
+
+	if err := json.Unmarshal(eventJSON, &event); err != nil {
+		return nil, err
+	}
+
+	unsignedJSON := event["unsigned"]
+
+	delete(event, "signatures")
+	delete(event, "unsigned")
```
Kegsay: Setting the precedent for using constants here would be a good idea IMO to prevent fat-fingering this. I appreciate we have good test coverage such that this would be immediately obvious, but I'm mainly thinking of the future when things inevitably get added. People will do what the surrounding code does, and if the surrounding code uses constants, so will they. This decreases the chance of them extracting the right key but then deleting a different key. The downside is that we're adding in a layer of indirection which may be unclear.

WDYT?

eventcrypto.go
```diff
@@ -0,0 +1,161 @@
...
+func AddContentHashesToEvent(eventJSON []byte) ([]byte, error) {
+	var event map[string]rawJSON
+
+	if err := json.Unmarshal(eventJSON, &event); err != nil {
+		return nil, err
+	}
+
+	unsignedJSON := event["unsigned"]
+
+	delete(event, "signatures")
```
Kegsay: You never add this back in to `event`. Why? You add `unsigned` back in, and work out `hashes` so this seems like a bug or a subtlety which needs explaining.

eventcrypto.go
```diff
@@ -0,0 +1,161 @@
...
+
+	if bytes.Compare(sha256Hash[:], []byte(hashes.Sha256)) != 0 {
+		return fmt.Errorf("Invalid Sha256 content hash: %v != %v", sha256Hash[:], []byte(hashes.Sha256))
+	}
+
+	return nil
+}
+
+// ReferenceSha256HashOfEvent returns the SHA-256 hash of the redacted event content.
+// This is used when refering to this event from other events.
```
Kegsay: > refering

eventcrypto.go
```diff
@@ -0,0 +1,161 @@
...
+}
+
+// SignEvent adds a ED25519 signature to the event for the given key.
+func SignEvent(signingName, keyID string, privateKey ed25519.PrivateKey, eventJSON []byte) ([]byte, error) {
+	redactedJSON, err := RedactEvent(eventJSON)
+	if err != nil {
+		return nil, err
+	}
+
+	signedJSON, err := SignJSON(signingName, keyID, privateKey, redactedJSON)
```
Kegsay: What is `signedJSON`? Is it the `redactedJSON` with a `signatures` key added (which we then parse out on `:139`)? It feels like we're doing a lot of back-and-forth (de)serialisation.

redactevent.go
NegativeMjark: Eh, it was one of the things I hated most about trying to read the synapse codebase fwiw.
It made it super hard to tell what was going on in the code because it meant that synapse was creating it's own names for everything rather than just using the strings from the matrix spec. But I agree that it would be nice if they were more resistant to being typo'd.

Kegsay: I wholeheartedly agree with the obfuscation it can cause, however the typo resistance outweighs this IMO. One can argue that provided we have enough tests then it's probably fine since typos in places like this will be flagged up by failing tests... I'm not sure which way to go on this. Whichever we decide we should be consistent though, and I think constants will be slightly better.

Kegsay:
Other than my outstanding comments, LGTM.

event.go
```diff
@@ -0,0 +1,225 @@
+package gomatrixserverlib
+
+import (
+	"encoding/json"
+	"fmt"
+	"golang.org/x/crypto/ed25519"
+	"time"
+)
+
+// An EventReference is a reference to a matrix event.
```
Kegsay: You can do better. What is the EventSHA256? Redacted hash or full event hash?

event.go
```diff
@@ -0,0 +1,225 @@
...
+	EventID     string
+	EventSHA256 Base64String
+}
+
+// An EventBuilder is used to build a new event.
+type EventBuilder struct {
+	// The user ID of the user sending the event.
+	Sender string `json:"sender"`
+	// The room ID of the room this event is in.
+	RoomID string `json:"room_id"`
```
Kegsay: So no presence events then?

event.go
```diff
@@ -0,0 +1,225 @@
...
+	// The type of the event.
+	EventType string `json:"type"`
```
Kegsay: As opposed to..? Why not just `Type`?

event.go
```diff
@@ -0,0 +1,225 @@
...
+	// The state_key of the event if the event is a state event or nil if the event is not a state event.
+	EventStateKey *string `json:"state_key,omitempty"`
```
Kegsay: As opposed to..? Why not just `StateKey`?

event.go
```diff
@@ -0,0 +1,225 @@
...
+	// The events that immediately preceeded this event in the room history.
+	PrevEvents []EventReference `json:"prev_events"`
+	// The events needed to authenticate this event.
+	AuthEvents []EventReference `json:"auth_events"`
+	// The event being redacted if this event is a "m.room.redaction".
+	Redacts  string `json:"redacts,omitempty"`
```
Kegsay: You mean event ID? It can't be "the event" entirely as a string right?

event.go
```diff
@@ -0,0 +1,225 @@
...
+
+// SetUnsigned sets the JSON unsigned key of the event.
+func (eb *EventBuilder) SetUnsigned(unsigned interface{}) (err error) {
+	eb.unsigned, err = json.Marshal(unsigned)
+	return
+}
+
+// An Event is a matrix event.
+// The event should always contain valid JSON.
+// If the event content hash is invalied then the event is redacted.
```
Kegsay: > invalied

event.go
```diff
@@ -0,0 +1,225 @@
...
+// Redacted events contain only the fields covered by the event signature.
+type Event struct {
+	redacted  bool
+	eventJSON []byte
+}
+
+var emptyEventReferenceList = []EventReference{}
```
Kegsay: Are you sure it's safe to do this? Won't appends modify this since this is just a global?

event.go
```diff
@@ -0,0 +1,225 @@
...
+
+// Build a new Event.
+// This is used when a local event is created on this server.
```
Kegsay: A tiny bit of clarification would be good to explain **when** this should be called: notably after you've set all your bits and pieces on the `EventBuilder`. Is it safe to build the same event multiple times? We should say either way.

event.go
```diff
@@ -0,0 +1,225 @@
...
+	}
+
+	result.eventJSON = eventJSON
+	return
+}
+
+// NewEventFromUntrustedJSON loads a new event from some JSON that may be invalid.
+// This checks that the event is valid JSON.
+// It also checks the content hashes to ensure the event has not been tampered with.
+// This should be used when receiving new events from remote servers.
```
Kegsay: ⭐️ Wonderful!

event.go
```diff
@@ -0,0 +1,225 @@
...
+
+// Redact returns a redacted copy of the event.
+func (e Event) Redact() Event {
+	if e.redacted {
+		return e
+	}
+	eventJSON, err := redactEvent(e.eventJSON)
+	if err != nil {
+		// This is unreachable for events created with EventBuilder.Build or NewEventFromUntrustedJSON
+		panic(fmt.Errorf("gomatrixserverlib: invalid event %v", err))
```
Kegsay: I would make this even longer and splat the `e.eventJSON` that was passed in so we have more context other than "ruh roh this isn't JSON" or whatever comes out of `json.Unmarshal`.

eventcrypto.go
```diff
@@ -140,6 +149,7 @@ func SignEvent(signingName, keyID string, privateKey ed25519.PrivateKey, eventJS
 		return nil, err
 	}
 
+	// Unmarshall the event JSON so that we can replace the signatures key.
```
Kegsay: > Unmarshall

Kegsay:
This PR is getting too large.

I'd like to see some tests for the `Event` struct to assert your proclamations for "This is unreachable..." and "This is reachable..." panics, but not in this PR.

event.go
NegativeMjark: appends don't modify the slice itself. (they can modify the backing array if there is capacity in the slice to do so.)

NegativeMjark:
> This PR is getting too large.

Would you be happier if I removed the event.go and did it as a separate PR?

Kegsay:
No, as I've already looked at it now :)

event.go
```diff
@@ -58,9 +60,10 @@ var emptyEventReferenceList = []EventReference{}
 
 // Build a new Event.
 // This is used when a local event is created on this server.
-func (eb *EventBuilder) Build(now time.Time, origin, keyID string, privateKey ed25519.PrivateKey) (result Event, err error) {
+func (eb *EventBuilder) Build(eventID string, now time.Time, origin, keyID string, privateKey ed25519.PrivateKey) (result Event, err error) {
```
Kegsay: You don't appear to use `eventID` anywhere.

Kegsay:
LGTM aside from comment.

Kegsay:
LGTM

PR #19, opened 2017-01-25
Move eventauth into gomatrixserverlib

NegativeMjark:
Modify eventauth to use gomatrixserverlib's Event type

### DIFF
--- a/event.go
+++ b/event.go
@@ -54,6 +54,19 @@ func (eb *EventBuilder) SetUnsigned(unsigned interface{}) (err error) {
 type Event struct {
 	redacted  bool
 	eventJSON []byte
+	fields    eventFields
+}
+
+type eventFields struct {
+	RoomID     string           `json:"room_id"`
+	EventID    string           `json:"event_id"`
+	Sender     string           `json:"sender"`
+	Type       string           `json:"type"`
+	StateKey   *string          `json:"state_key"`
+	Content    rawJSON          `json:"content"`
+	PrevEvents []EventReference `json:"prev_events"`
+	AuthEvents []EventReference `json:"auth_events"`
+	Redacts    string           `json:"redacts"`
 }
 
 var emptyEventReferenceList = []EventReference{}
@@ -106,6 +119,7 @@ func (eb *EventBuilder) Build(eventID string, now time.Time, origin, keyID strin
 	}
 
 	result.eventJSON = eventJSON
+	err = json.Unmarshal(eventJSON, &result.fields)
 	return
 }
 
@@ -145,14 +159,18 @@ func NewEventFromUntrustedJSON(eventJSON []byte) (result Event, err error) {
 	}
 
 	result.eventJSON = eventJSON
+	err = json.Unmarshal(eventJSON, &result.fields)
 	return
 }
 
 // NewEventFromTrustedJSON loads a new event from some JSON that must be valid.
 // This will be more efficient than NewEventFromUntrustedJSON since it can skip cryptographic checks.
 // This can be used when loading matrix events from a local database.
-func NewEventFromTrustedJSON(eventJSON []byte, redacted bool) Event {
-	return Event{redacted, eventJSON}
+func NewEventFromTrustedJSON(eventJSON []byte, redacted bool) (result Event, err error) {
+	result.redacted = redacted
+	result.eventJSON = eventJSON
+	err = json.Unmarshal(eventJSON, &result.fields)
+	return
 }
 
 // Redacted returns whether the event is redacted.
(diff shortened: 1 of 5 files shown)


eventauth_test.go
```diff
@@ -393,17 +413,6 @@ func TestAllowedFirstJoin(t *testing.T) {
...
 			"type": "m.room.member",
 			"state_key": "@u1:a",
-			"sender": "@u1:a",
-			"room_id": "!r1:a",
-			"event_id": "$e4:a",
-			"content": {"membership": "join"},
-			"prev_events": [[0, {}]],
-			"unsigned": {
-				"not_allowed": "The prev_events event ID is not a string"
-			}
```
NegativeMjark: I can't actually use this test case anymore because the prev_events are parsed inside the constructor for events rather than inside the auth checks.

event.go
```diff
@@ -230,3 +248,89 @@ func (e Event) KeyIDs(signingName string) []string {
...
 	return verifyEventSignature(signingName, keyID, publicKey, e.eventJSON)
 }
+
+// IsState returns true if the event is a state event.
+func (e Event) IsState() bool {
+	return e.fields.StateKey != nil
+}
+
+// StateKey returns the "state_key" of the event, or the empty string if it's
+func (e Event) StateKey() string {
```
Kegsay: Comment isn't finished. Also, it feels deeply wrong to be defaulting to the empty string when the empty string is used so much as a state key. I feel this buys us nothing, since you need to know if this event is a state event in order for this to make any sense, so the typical use will be:
```go
if e.IsState() {
    key := e.StateKey()
}
```

...at which point you might as well be doing the `nil` deference yourself. What exactly is the intended use case here?

event.go
```diff
@@ -230,3 +248,89 @@ func (e Event) KeyIDs(signingName string) []string {
...
+}
+
+// UnmarshalJSON implements json.Unmarshaller
+func (er *EventReference) UnmarshalJSON(data []byte) error {
+	var tuple []rawJSON
+	if err := json.Unmarshal(data, &tuple); err != nil {
+		return err
+	}
+	if len(tuple) != 2 {
+		return fmt.Errorf("gomatrixserverlib: invalid event reference")
```
Kegsay: Be more helpful. If we get this in the wild, whoever is debugging this is going to be really sad that there isn't a more informative error message: state the length of the tuple and the expected length like we do for tests?

event.go
```diff
@@ -230,3 +248,89 @@ func (e Event) KeyIDs(signingName string) []string {
...
+	return nil
+}
+
+// MarshalJSON implements json.Marshaller
+func (er EventReference) MarshalJSON() ([]byte, error) {
+	hashes := struct {
+		SHA256 Base64String `json:"sha256"`
+	}{er.EventSHA256}
+
+	tuple := []interface{}{er.EventID, hashes}
```
Kegsay: It feels messy that we don't seem to have a well-formed struct for references and instead consistently refer to them as "tuples" and indexing into them rather than treating them as 2-field structs. How practical is it to make this a bit more structured (given the serialised JSON form isn't an object).

eventauth.go
```diff
@@ -204,25 +191,25 @@ func Allowed(event Event, authEvents AuthEvents) error {
 // createEventAllowed checks whether the m.room.create event is allowed.
 // It returns an error if the event is not allowed.
 func createEventAllowed(event Event) error {
-	if event.StateKey == nil {
+	if !event.IsState() {
 		return errorf("create event missing state key")
 	}
-	if *event.StateKey != "" {
-		return errorf("create event state key is not empty: %q", event.StateKey)
+	if event.StateKey() != "" {
```
Kegsay: I feel really uncomfortable with this given the default semantics if there is no state key. We're relying completely that we have checked `IsState()` prior to calling this: if we (re)move that then this check will NOT fail. This is really subtle: it's clearer in the previous diff because it's obvious there is a correlation since you're doing explicit `nil` checks on the same field.

I'd rather we removed `IsState()` and only exposed a nullable string return value for `StateKey()` to make this completely obvious.

Kegsay: Or alternatively expose a `StateKeyEquals(str)` function which will handle the non-existent case for you, since most of the time you are doing comparisons.

eventcontent.go
```diff
@@ -175,7 +175,7 @@ func (c *powerLevelContent) userLevel(userID string) int64 {
 }
 
 // eventLevel returns the power level needed to send an event in the room.
-func (c *powerLevelContent) eventLevel(eventType string, eventStateKey *string) int64 {
+func (c *powerLevelContent) eventLevel(eventType string, isState bool) int64 {
```
Kegsay: It's not obvious at call-sites what `eventLevel(eventType, false)` means. Can we enum this please to make it obvious to callers please?

NegativeMjark: Does ac667b7 work for you?

Kegsay: WFM!

Kegsay:
LGTM

PR #20, opened 2017-01-27
Implement the state resolution algorithm

NegativeMjark:


event.go
```diff
@@ -45,7 +45,9 @@ type EventBuilder struct {
 	// The events needed to authenticate this event.
 	AuthEvents []EventReference `json:"auth_events"`
 	// The event ID of the event being redacted if this event is a "m.room.redaction".
-	Redacts  string `json:"redacts,omitempty"`
+	Redacts string `json:"redacts,omitempty"`
+	// The depth of the event.
```
Kegsay: It might be nice to state what this depth relates to? Depth from the `m.room.create` event?

stateresolution.go
```diff
@@ -0,0 +1,255 @@
...
+	return r.resolvedMembers[key], nil
+}
+
+func (r *stateResolver) addConflicted(events []Event) {
+	type conflictKey struct {
+		eventType string
+		stateKey  string
+	}
+	offsets := map[conflictKey]int{}
+	// Split up the conflicted events into blocks with the same state key.
```
Kegsay: I assume by "block" you mean "list"? And I assume by "the same state key" you mean "the same event type and state key"?

Superficially it sounds like you're taking a list of events and putting them into a dict keyed off the `conflictKey` (type/state key tuple), but I must be missing something because you do Magic depending on the event type. Needs much more explanation for me to grok :(

stateresolution.go
```diff
@@ -0,0 +1,255 @@
...
+	for _, event := range events {
+		key := conflictKey{event.Type(), *event.StateKey()}
+		// Work out which block to add the event to.
+		var block *[]Event
+		// By default we add the event to a block in the others list.
+		blockList := &r.others
```
Kegsay: :confused: `blockList` makes it sound like "a list of things which should be blocked"? But then you also have "block" as in "chunk/group". This is making me confused as to what the point is for these "blocks".

stateresolution.go
```diff
@@ -0,0 +1,255 @@
...
+		switch key.eventType {
+		case "m.room.create":
+			if key.stateKey == "" {
+				block = &r.creates
+			}
+		case "m.room.power_levels":
+			if key.stateKey == "" {
+				block = &r.powerLevels
```
Kegsay: Right, I think I know what you're doing, and I think this function is needlessly complicated considering what it's doing. I think it's confusing because you're desperately trying to only have 1 place where you append to the list, so you go to great lengths to end up assigning `block` to the right `[]Event` which obfuscates what you're actually doing, which really does appear to be just grouping events into pre-defined lists based on the `conflictKey`.

Look at `:108` for a good example of why I think this is confusing. If `block == nil` it also implies that you need to look at `blockList`, but you can't do `if blockList != nil` because you have already defaulted it to the `others` list.

The entire point is to "do checks, assign to list, continue" and I would prefer if the code reflected that better e.g:

```go
for _, event := range events {
    if event.Type() == "m.room.create" && *event.StateKey() == "" {
        r.creates = append(r.creates, event)
        continue
    }
    // etc...
}
```

I don't understand why you store events where there's custom state keys (e.g member events) as a 2d array rather than a map which feels more natural. Why do you do that? If you didn't, you wouldn't need to maintain this `offsets` structure.

stateresolution.go
```diff
@@ -0,0 +1,255 @@
...
+			}
+			// Get the address of the block in the block list.
+			block = &(*blockList)[offset]
+		}
+		// Add the event to the block.
+		*block = append(*block, event)
+	}
+}
+
+// Add an event to the resolved auth events.
```
Kegsay: I don't understand the significance of the term "resolved". AFAIK, auth events are a select number of state event types (the ones inside the switch in the body of this function) and that's it. There is no concept of them being "resolved" or "unresolved" or anything else. Indeed, grepping for `unresolved` turns up nothing, so what does this mean?

stateresolution.go
```diff
@@ -0,0 +1,255 @@
...
+		r.resolvedCreate = event
+	case "m.room.power_levels":
+		r.resolvedPowerLevels = event
+	case "m.room.join_rules":
+		r.resolvedJoinRules = event
+	case "m.room.member":
+		r.resolvedMembers[*event.StateKey()] = event
+	case "m.room.third_party_invite":
+		r.resolvedThirdPartyInvites[*event.StateKey()] = event
+	}
```
Kegsay: And if the event type isn't any of those? That feels like a logic error and I feel we should panic. We should never encounter that. I'd rather we failed fast rather than silently continued like we do currently.

stateresolution.go
```diff
@@ -0,0 +1,255 @@
...
+	case "m.room.third_party_invite":
+		r.resolvedThirdPartyInvites[stateKey] = nil
+	}
+}
+
+// resolveAndAddAuthBlocks resolves each block of conflicting auth state events in a list of blocks
+// where all the blocks have the same event type.
+// Once every block has been resolved the resulting events are added to the events used for auth checks.
+func (r *stateResolver) resolveAndAddAuthBlocks(blocks [][]Event) {
+	start := len(r.result)
```
Kegsay: Given we blindly append to `result`, this makes these calls presumably unsafe to be called multiple times?

stateresolution.go
```diff
@@ -0,0 +1,255 @@
...
+	joinRules                 []Event
+	thirdPartyInvites         [][]Event
+	members                   [][]Event
+	others                    [][]Event
+	resolvedCreate            *Event
+	resolvedPowerLevels       *Event
+	resolvedJoinRules         *Event
+	resolvedThirdPartyInvites map[string]*Event
+	resolvedMembers           map[string]*Event
+	result                    []Event
```
Kegsay: Are there any kind of guarantees made as to what `result` will contain? The API exposed by `stateResolver` implies the caller knows how this works and when `result` is nonsensical: I'm pretty sure you need to call `resolveAndAddAuthBlocks` for each event type for this to actually have the right result, and if you call `resolveAndAddAuthBlocks` too many times it'll also be incorrect.

stateresolution.go
NegativeMjark: "resolved" means that any conflict has been resolved.

stateresolution.go
```diff
@@ -80,43 +80,45 @@ func (r *stateResolver) addConflicted(events []Event) {
 		stateKey  string
 	}
 	offsets := map[conflictKey]int{}
-	// Split up the conflicted events into blocks with the same state key.
+	// Split up the conflicted events into blocks with the same type and state key.
+	// Separate the auth events into sepecially named lists because they have
```
Kegsay: > sepecially

Other than that, this is a good explanation for why you're splitting them, thank you!

stateresolution.go
```diff
@@ -0,0 +1,257 @@
...
+// Once every block has been resolved the resulting events are added to the events used for auth checks.
+func (r *stateResolver) resolveAndAddAuthBlocks(blocks [][]Event) {
+	start := len(r.result)
+	for _, block := range blocks {
+		if event := r.resolveAuthBlock(block); event != nil {
+			r.result = append(r.result, *event)
+		}
+	}
+	for i := start; i < len(r.result); i++ {
+		r.addAuthEvent(&r.result[i])
```
Kegsay: Uhh, why aren't you just doing this when you append the event to the list on `:166`? I bet it's probably because if you do that it might change how the auth block is resolved so it's important that it's done after or something like that, but there's not a single comment to tell me this.

stateresolution.go
```diff
@@ -0,0 +1,257 @@
...
+			return event
+		}
+	}
+	return block[0].event
+}
+
+// A conflictedEvent is used to sort the events in a block by depth and sha1 of event ID.
+type conflictedEvent struct {
+	depth       int64
+	eventIDSHA1 [sha1.Size]byte
```
Kegsay: Excuse my ignorance, by what is the value in SHA1ing the event ID? Why can't we just use the event ID? If I had to guess: something about not being able to rig the system and make your event be preferred in the event of a conflict at the same depth? That kinda feels pointless though since it just adds a rather trivial extra step (you need to tweak your event ID a few times to get it to sort earlier/later)? Comments on why we SHA1 would be superb.

stateresolution.go
```diff
@@ -0,0 +1,257 @@
...
+	event       *Event
+}
+
+// A conflictedEventSorter is used to sort the events using sort.Sort.
+type conflictedEventSorter []conflictedEvent
```
Kegsay: :heart: this is nice. However, I have no confidence your `Less(i j int)` code is correct. Can you add some tests to convince me please?

Kegsay: You should also state the sort order (ASC / DESC)

stateresolution.go
```diff
@@ -0,0 +1,279 @@
...
+	result := block[0].event
+	// Temporarily add the candidate event to the auth events.
+	r.addAuthEvent(result)
+	for i := 1; i < len(block); i++ {
+		event := block[i].event
+		// Check if the next event passes authentication checks against the current candidate.
+		if Allowed(*event, r) == nil {
+			// If the event passes authentication checks pick it as the current candidate.
+			result = event
+			r.addAuthEvent(result)
```
Kegsay: Eurgh, this is confusing. Can you please state the rules in a block comment somewhere, otherwise it just makes it look like this has errors in it. From what I can tell:
- if there's only 1 candidate event, choose it (would be nice if we bailed early rather than the subtle `i := 1` skipover)
- else there is >1 candidate event, does it pass auth checks?:
    * YES: It becomes the new candidate. Go to next event.
    * NO: Return the last candidate which passed auth checks.

In all cases, `stateResolver` needs to be "reset" to the previous auth event set. Is this even remotely correct?

stateresolution.go
```diff
@@ -0,0 +1,279 @@
...
+	block := make([]conflictedEvent, len(events))
+	for i := range events {
+		event := &events[i]
+		block[i] = conflictedEvent{
+			depth:       event.Depth(),
+			eventIDSHA1: sha1.Sum([]byte(event.EventID())),
+			event:       event,
+		}
+	}
+	sort.Sort(conflictedEventSorter(block))
```
Kegsay: `:234~:243` are an exact duplicate of code in `resolveAuthBlock`, and it performs a single well-defined task (sorting events by depth and SHA1 of the event ID). Can we factor this out into:
```
sortByDepth(events []Event) []conflictedEvent
```
please?

stateresolution.go
```diff
@@ -0,0 +1,279 @@
...
+
+	// Start at the "newest" event, that is the one with the highest depth, and go
+	// backward through the list until we find one that passes authentication checks.
```
Kegsay: Interesting. Why do we do this by the newest event but the other stuff was by the oldest event? I'm guessing that it has something to do with "well it doesn't affect the integrity of the room so just let everyone clobber everything"? It feels like a good place to mention the intent here.

stateresolution.go
```diff
@@ -0,0 +1,279 @@
...
+	for i := len(block) - 1; i > 0; i-- {
+		event := block[i].event
+		if Allowed(*event, r) == nil {
+			return event
+		}
+	}
+	return block[0].event
```
Kegsay: You seem to like the style of skipping over one element in an array to do checks then default to returning the "single element" case. I don't really like it. Off-by-one errors are frequent in a lot of code and we need not exacerbate it by being clever with our indexing IMO when the following suffices:
```go
if len(block) == 1 {
  return block[0].event
}
// handle multi-element case.
```

stateresolution.go
```diff
@@ -41,12 +41,17 @@ func ResolveStateConflicts(conflicted []Event, authEvents []Event) []Event {
...
-	creates                   []Event
-	powerLevels               []Event
-	joinRules                 []Event
-	thirdPartyInvites         [][]Event
-	members                   [][]Event
-	others                    [][]Event
+	// Lists of lists of events to resolve grouped by event type and state key:
+	//   * creates, powerLevels, joinRules have empty state keys.
+	//   * members and thirdPartyInvites are grouped by state key.
+	//   * the others are grouped by type and state key.
```
Kegsay: The `others` then feels like it should be a 3d-array based on this description but it isn't, presumably because the "key" is the tuple of (event type, state key) already. Perhaps word it as:
```go
// List of lists of events to resolve grouped by the tuple of (event type, state key):
// * creates, powerLevels and joinRules have 0-length state keys so don't
//   need to be grouped by the tuple.
// * members, thirdPartyInvites and others are grouped by the tuple: each
//   element in the outer array represents a specific tuple, and each Event in the inner
//   array is guaranteed to have the same type/state key as each other.
```

stateresolution.go
NegativeMjark: Done 225fa7e

Kegsay:
Other than tests for the sorting function and a comment reword https://github.com/matrix-org/gomatrixserverlib/pull/20#discussion_r98223983 , LGTM!

NegativeMjark:
@Kegsay I've added a test for the sort order, and a reworded the comment for the `other [][]Event` field

Kegsay:
LGTM

stateresolution.go
ara4n: might be worth spelling that out in the comment then

PR #26, opened 2017-03-10
s/AuthEvents/AuthEventProvider/g and add map-based impl

Kegsay:
With tests.

### DIFF
--- a/eventauth.go
+++ b/eventauth.go
@@ -160,8 +160,8 @@ func thirdPartyInviteToken(thirdPartyInviteData rawJSON) (string, error) {
 	return thirdPartyInvite.Signed.Token, nil
 }
 
-// AuthEvents are the state events needed to authenticate an event.
-type AuthEvents interface {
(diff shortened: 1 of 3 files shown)


eventauth.go
```diff
@@ -175,6 +175,61 @@ type AuthEvents interface {
 	ThirdPartyInvite(stateKey string) (*Event, error)
 }
 
+type stateKeyTuple struct {
+	Type     string
+	StateKey string
+}
+
+// AuthEventProviderMap is an implementation of AuthEventProvider backed by a map.
+type AuthEventProviderMap struct {
```
NegativeMjark: Why not just call this `AuthEvents`? AuthEventProviderMap sounds like a map of AuthEventProviders to me.

Kegsay: WFM

Kegsay: Done. https://github.com/matrix-org/gomatrixserverlib/pull/26/commits/128b665d1d4e1355c777e7fe8e1745d5c53b70de

PR #28, opened 2017-03-17
Add StateKeyTuple type. Add helper functions for auth.

Kegsay:
Helper functions are designed to map easily between StateNeeded,
StateKeyTuples and []EventReference (auth_events).

### DIFF
--- a/eventauth.go
+++ b/eventauth.go
@@ -44,6 +44,69 @@ type StateNeeded struct {
 	ThirdPartyInvite []string
 }
 
+// Tuples returns the needed state key tuples for performing auth on an event.
+func (s StateNeeded) Tuples() (res []StateKeyTuple) {
+	if s.Create {
+		res = append(res, StateKeyTuple{"m.room.create", ""})
+	}
+	if s.JoinRules {
+		res = append(res, StateKeyTuple{"m.room.join_rules", ""})
+	}
+	if s.PowerLevels {
+		res = append(res, StateKeyTuple{"m.room.power_levels", ""})
+	}
+	for _, userID := range s.Member {
+		res = append(res, StateKeyTuple{"m.room.member", userID})
+	}
+	for _, token := range s.ThirdPartyInvite {
+		res = append(res, StateKeyTuple{"m.room.third_party_invite", token})
+	}
+	return
+}
+
+// AuthEventReferences returns the auth_events references for the StateNeeded. Returns an error if the
+// provider returns an error. If an event is missing from the provider but is required in StateNeeded, it
+// is skipped under the assumption that the event will fail authentication checks.
+func (s StateNeeded) AuthEventReferences(provider AuthEventProvider) (refs []EventReference, err error) {
+	var e *Event
+	if s.Create {
+		if e, err = provider.Create(); err != nil {
+			return
+		} else if e != nil {
+			refs = append(refs, e.EventReference())
+		}
+	}
+	if s.JoinRules {
+		if e, err = provider.JoinRules(); err != nil {
+			return
+		} else if e != nil {
+			refs = append(refs, e.EventReference())
+		}
+	}
+	if s.PowerLevels {
+		if e, err = provider.PowerLevels(); err != nil {
+			return
+		} else if e != nil {
+			refs = append(refs, e.EventReference())
+		}
+	}
+	for _, userID := range s.Member {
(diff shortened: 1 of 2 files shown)


eventauth.go
```diff
@@ -86,6 +86,40 @@ func StateNeededForAuth(events []Event) (result StateNeeded) {
...
+}
+
+// AuthEventReferencesFromStateNeeded returns a list of event references which can be used under the 'auth_events' key
+// of a matrix event. 'authEvents' should contain all the current state events which may be used for auth, as this
+// function may require their event reference.
+func AuthEventReferencesFromStateNeeded(authEvents map[StateKeyTuple]*Event, needed StateNeeded) (refs []EventReference) {
+	tuples := TuplesFromStateNeeded(needed)
+	for _, t := range tuples {
+		e, ok := authEvents[t]
+		if ok && e != nil {
```
NegativeMjark: Technically the `ok` check is unnecessary here because `e` will be `nil` if the tuple isn't present in the map.

Kegsay: Done. https://github.com/matrix-org/gomatrixserverlib/pull/28/commits/56239aa496b880a7ee4b7f4ee11cc72ffb4b70a3

eventauth.go
```diff
@@ -86,6 +86,40 @@ func StateNeededForAuth(events []Event) (result StateNeeded) {
...
+	for _, token := range needed.ThirdPartyInvite {
+		res = append(res, StateKeyTuple{"m.room.third_party_invite", token})
+	}
+	return
+}
+
+// AuthEventReferencesFromStateNeeded returns a list of event references which can be used under the 'auth_events' key
+// of a matrix event. 'authEvents' should contain all the current state events which may be used for auth, as this
+// function may require their event reference.
+func AuthEventReferencesFromStateNeeded(authEvents map[StateKeyTuple]*Event, needed StateNeeded) (refs []EventReference) {
```
NegativeMjark: I'd be happier if this was a method on `StateNeeded` that took a `AuthEventsProvider` interface.

eventauth.go
```diff
@@ -86,6 +86,40 @@ func StateNeededForAuth(events []Event) (result StateNeeded) {
 	return
 }
 
+// TuplesFromStateNeeded returns the needed state key tuples for performing auth on an event.
+func TuplesFromStateNeeded(needed StateNeeded) (res []StateKeyTuple) {
```
NegativeMjark: It might be neater to make this a method on `StateNeeded`.

eventauth.go
NegativeMjark: Something like this maybe?:

```go
func (needed StateNeeded) AuthEventReferences(authEvents AuthEventProvider) (refs []EventReference, err error) {
   var *event Event
   if needed.Create {
      if event, err = authEvents.Create(); err != nil {
           return
      } else if (event != nil) {
           append(refs, event.EventReference())
      }
   }
   ...
}
```

eventauth.go
Kegsay: Done. https://github.com/matrix-org/gomatrixserverlib/pull/28/commits/93b606350db0e699bd9ab4be57d158a9cdddce72

eventauth.go
Kegsay: SGTM. Done. https://github.com/matrix-org/gomatrixserverlib/pull/28/commits/93b606350db0e699bd9ab4be57d158a9cdddce72

eventauth.go
```diff
@@ -44,6 +44,69 @@ type StateNeeded struct {
...
+	}
+	for _, token := range s.ThirdPartyInvite {
+		res = append(res, StateKeyTuple{"m.room.third_party_invite", token})
+	}
+	return
+}
+
+// AuthEventReferences returns the auth_events references for the StateNeeded. Returns an error if the
+// provider returns an error. If an event is missing from the provider but is required in StateNeeded, it
+// is skipped under the assumption that the event will fail authentication checks.
```
NegativeMjark: Strictly speaking this is misleading. An event may still pass the authentication checks even though a tuple is missing from StateNeeded. For example `m.room.member` events need the previous value of the `m.room.member` event to auth, but when a user joins the room for the first time there is no previous event.

Kegsay: Tweaked comment: https://github.com/matrix-org/gomatrixserverlib/pull/28/commits/121fba145760eda2b58e95604f0e9e720b08da5f

PR #29, opened 2017-04-13
Add ClientEvent and more Event functions

Kegsay:
This is required for the sync server as per https://github.com/matrix-org/dendrite/pull/69

### DIFF
--- a/clientevent.go
+++ b/clientevent.go
@@ -15,6 +15,17 @@
 
 package gomatrixserverlib
 
+// EventFormat specifies the format of a client event
+type EventFormat int
+
+const (
+	// FormatAll will include all client event keys
+	FormatAll EventFormat = iota
+	// FormatSync will include only the event keys required by the /sync API. Notably, this
+	// means the 'room_id' will be missing from the events.
+	FormatSync
+)
+
 // ClientEvent is an event which is fit for consumption by clients, in accordance with the specification.
 type ClientEvent struct {
 	Content        rawJSON `json:"content"`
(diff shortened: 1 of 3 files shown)


clientevent.go
```diff
@@ -0,0 +1,53 @@
...
+	"encoding/json"
+)
+
+// ClientEvent is an event which is fit for consumption by clients, in accordance with the specification.
+type ClientEvent struct {
+	Content        json.RawMessage `json:"content"`
+	Sender         string          `json:"sender"`
+	Type           string          `json:"type"`
+	StateKey       *string         `json:"state_key,omitempty"`
+	Unsigned       json.RawMessage `json:"unsigned,omitempty"`
```
NegativeMjark: Does json.RawMessage encode correctly? Or do you end up with the base64 encoded bytes instead?

clientevent.go
```diff
@@ -0,0 +1,53 @@
...
+ */
+
+package gomatrixserverlib
+
+import (
+	"encoding/json"
+)
+
+// ClientEvent is an event which is fit for consumption by clients, in accordance with the specification.
+type ClientEvent struct {
```
NegativeMjark: Is is useful for ClientEvent to be a structure, or could it just be a RawJSON message containing the JSON encoded bytes?

Kegsay: At present, it isn't useful for it to be a `struct` as opposed to a thing which can be marshalled into valid JSON. I don't mind which it is.

Kegsay: EDIT: Scratch that, it *is* useful because we situationally want to discard the `room_id` key (/sync doesn't have it set since it's in a higher up key, but other APIs do expose the `room_id` key when the context does not exist e.g AS API).

clientevent.go
Kegsay: ... This language.... Fixed. https://github.com/matrix-org/gomatrixserverlib/pull/29/commits/6259d600d934c93dcb85cadddcbcdb378a24a92b

clientevent.go
```diff
@@ -15,39 +15,43 @@
...
+	StateKey *string `json:"state_key,omitempty"`
+	Type     string  `json:"type"`
+	Unsigned rawJSON `json:"unsigned,omitempty"`
 }
 
-// ToClientEvents converts server events to client events
-func ToClientEvents(serverEvs []Event) []ClientEvent {
+// ToClientEvents converts server events to client events. If omitRoomIDs is true, the room_id will not be
+// set on the resulting events.
+func ToClientEvents(serverEvs []Event, omitRoomIDs bool) []ClientEvent {
```
NegativeMjark: Not sure I like the omitRoomIDs being a boolean. It might be nicer to make it an enum.

NegativeMjark: Also the double negative is a bit jarring. Maybe `includeRoomIDs` would be a better idea.

PR #31, opened 2017-04-24
Add a KeyRing for storing server keys for verifying signed JSON

NegativeMjark:
This implements the logic in https://github.com/matrix-org/synapse/blob/master/synapse/crypto/keyring.py for checking signed JSON by fetching the ed25519 keys from the remote server either directly or via a perspective server.

### DIFF
--- a/client.go
+++ b/client.go
@@ -153,16 +153,20 @@ func (fc *Client) ServerKeys(
 		Host:   matrixServer,
 		Path:   "/_matrix/key/v2/query",
 	}
+
+	// The request format is:
+	// { "server_keys": { "<server_name>": { "<key_id>": { "minimum_valid_until_ts": <ts> }}}
 	type keyreq struct {
 		MinimumValidUntilTS Timestamp `json:"minimum_valid_until_ts"`
 	}
-	request := map[string]map[string]keyreq{}
-
+	request := struct {
+		ServerKeys map[string]map[string]keyreq `json:"server_keys"`
+	}{map[string]map[string]keyreq{}}
 	for k, ts := range keyRequests {
-		server := request[k.ServerName]
+		server := request.ServerKeys[k.ServerName]
 		if server == nil {
 			server = map[string]keyreq{}
-			request[k.ServerName] = server
+			request.ServerKeys[k.ServerName] = server
 		}
 		server[k.KeyID] = keyreq{ts}
 	}
@@ -188,15 +192,20 @@ func (fc *Client) ServerKeys(
 		return nil, fmt.Errorf("HTTP %d : %s", response.StatusCode, errorOutput)
 	}
 
-	var body map[string]map[string]ServerKeys
+	var body struct {
+		ServerKeys []ServerKeys `json:"server_keys"`
+	}
 	if err = json.NewDecoder(response.Body).Decode(&body); err != nil {
 		return nil, err
 	}
 
 	result := map[PublicKeyRequest]ServerKeys{}
-	for server, keys := range body {
-		for keyID, keyData := range keys {
-			result[PublicKeyRequest{server, keyID}] = keyData
+	for _, keys := range body.ServerKeys {
+		for keyID := range keys.VerifyKeys {
+			result[PublicKeyRequest{keys.ServerName, keyID}] = keys
+		}
+		for keyID := range keys.OldVerifyKeys {
+			result[PublicKeyRequest{keys.ServerName, keyID}] = keys
 		}
 	}
 	return result, nil
--- a/client.go
+++ b/client.go
@@ -144,6 +144,11 @@ func (fc *Client) LookupUserInfo(matrixServer, token string) (u UserInfo, err er
 }
 
 // LookupServerKeys lookups up the keys for a matrix server from a matrix server.
+// Takes a map from (server name, key ID) pairs to timestamps.
+// The timestamps tell the server when the keys need to be valid until.
+// Perspective servers can use that timestamp to determine whether they can
+// return a cached copy of the keys or whether they will need to retrieve a fresh
+// copy of the keys.
 // Returns the keys or an error if there was a problem talking to the server.
 func (fc *Client) LookupServerKeys(
 	matrixServer string, keyRequests map[PublicKeyRequest]Timestamp,
--- a/client.go
+++ b/client.go
@@ -144,7 +144,9 @@ func (fc *Client) LookupUserInfo(matrixServer, token string) (u UserInfo, err er
 }
 
 // LookupServerKeys lookups up the keys for a matrix server from a matrix server.
-// Takes a map from (server name, key ID) pairs to timestamps.
+// The first argument is the name of the matrix server to download the keys from.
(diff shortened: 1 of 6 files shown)


keyring_test.go
```diff
@@ -0,0 +1,139 @@
...
+		t.Fatal(err)
+	}
+	if len(results) != 1 || results[0].Result == nil {
+		t.Fatalf("VerifyJSON(): Wanted [{Result: <some error>}] got %#v", results)
+	}
+}
+
+func TestVerifyJSONsDistantFutureFails(t *testing.T) {
+	// Check that trying to verify JSON from the distant future fails.
+	distantFuture := Timestamp(2000000000000)
```
richvdh: o/` the humans are dead.

keyring.go
```diff
@@ -0,0 +1,303 @@
...
+	k.checkUsingKeys(requests, results, keyIDs, keysFromDatabase)
+
+	for i := range k.KeyFetchers {
+		keyRequests := k.publicKeyRequests(requests, results, keyIDs)
+		if len(keyRequests) == 0 {
+			// There aren't any keys to fetch so we can stop here.
+			// This means that we've checked every JSON object we can check.
+			return results, nil
+		}
+		keysFetched, err := k.KeyFetchers[i].FetchKeys(keyRequests)
```
richvdh: do we need to worry about the situation where we have several calls to `VerifyJSON` in parallel, and then end up doing several fetches of the same key rather than just waiting for the first to complete?

keyring.go
```diff
@@ -0,0 +1,303 @@
...
+		}
+	}
+}
+
+// A PerspectiveKeyFetcher fetches server keys from a single perspective server.
+type PerspectiveKeyFetcher struct {
+	// The name of the perspective server to fetch keys from.
+	ServerName string
+	// The ed25519 public keys the perspective server must sign responses with.
+	ServerKeys map[string]ed25519.PublicKey
```
richvdh: oh good, something else called ServerKeys :-1:

keyring.go
```diff
@@ -0,0 +1,303 @@
...
+	results, err := d.Client.ServerKeys(serverName, requests)
+	if err != nil {
+		return nil, err
+	}
+
+	for req, keys := range results {
+		// Check that the keys are valid for the server.
+		checks, _, _ := CheckKeys(req.ServerName, time.Unix(0, 0), keys, nil)
+		if !checks.AllChecksOK {
+			// This is bad because it means that the perspective server was trying to feed us an invalid response.
```
richvdh: no perspective server here

keys.go
```diff
@@ -31,21 +30,52 @@ import (
...
 
+// UnmarshalJSON implements json.Unmarshaler
+func (keys *ServerKeys) UnmarshalJSON(data []byte) error {
+	keys.Raw = data
+	return json.Unmarshal(data, &keys.ServerKeyFields)
+}
+
+// MarshalJSON implements json.Marshaler
+func (keys ServerKeys) MarshalJSON() ([]byte, error) {
+	return keys.Raw, nil
```
richvdh: this looks... surprising

client.go
```diff
@@ -141,3 +142,74 @@ func (fc *Client) LookupUserInfo(matrixServer, token string) (u UserInfo, err er
 
 	return
 }
+
+// ServerKeys lookups up the keys for a matrix server from a matrix server.
+// Returns the keys or a error if there was a problem talking to
```
NegativeMjark: Fixed d6f9dc6

client.go
```diff
@@ -141,3 +142,74 @@ func (fc *Client) LookupUserInfo(matrixServer, token string) (u UserInfo, err er
...
+func (fc *Client) ServerKeys(
```
NegativeMjark: Done d6f9dc6

client.go
```diff
@@ -141,3 +142,74 @@ func (fc *Client) LookupUserInfo(matrixServer, token string) (u UserInfo, err er
...
+	matrixServer string, keyRequests map[PublicKeyRequest]Timestamp,
```
NegativeMjark: Done a8780a8

NegativeMjark: Done 04ef89b

keyring.go
```diff
@@ -0,0 +1,303 @@
...
+
+import (
+	"fmt"
+	"golang.org/x/crypto/ed25519"
+	"strings"
+	"time"
+)
+
+// A Timestamp is a millisecond posix timestamp.
+type Timestamp uint64
```
NegativeMjark: Done 7ba7714

keyring.go
```diff
@@ -0,0 +1,303 @@
...
+type PublicKeyRequest struct {
+	// The server to fetch a key for.
+	ServerName string
+	// The ID of the key to fetch.
+	KeyID string
+}
+
+// A KeyFetcher is a way of fetching public keys in bulk.
+type KeyFetcher interface {
+	// Lookup a batch of public keys.
```
NegativeMjark: Done 6251765

keyring.go
NegativeMjark: Fixed 02571d4

keyring.go
NegativeMjark: Fixed d6f9dc6

keys.go
NegativeMjark: Does this comment help? 192dcd5

keyring.go
```diff
@@ -0,0 +1,303 @@
...
+	FetchKeys(requests map[PublicKeyRequest]Timestamp) (map[PublicKeyRequest]ServerKeys, error)
+}
+
+// A KeyDatabase is a store for caching public keys.
+type KeyDatabase interface {
+	KeyFetcher
+	// Add a block public keys to the database.
```
NegativeMjark: Done 6251765. I don't know if one follows the `<name of method>... does` for interfaces.

keyring.go
NegativeMjark: Maybe, but the levels of synchronisation between goroutines, and possibly processes needed to prevent that is quite saddening. I've left a TODO e8de828

NegativeMjark:
@richvdh I think I've responded to the review comments PTAL

client.go
```diff
@@ -143,9 +143,16 @@ func (fc *Client) LookupUserInfo(matrixServer, token string) (u UserInfo, err er
 	return
 }
 
-// ServerKeys lookups up the keys for a matrix server from a matrix server.
-// Returns the keys or a error if there was a problem talking to
-func (fc *Client) ServerKeys(
+// LookupServerKeys lookups up the keys for a matrix server from a matrix server.
```
richvdh: lookups up

keys.go
```diff
@@ -62,6 +62,7 @@ func (keys *ServerKeys) UnmarshalJSON(data []byte) error {
 
 // MarshalJSON implements json.Marshaler
 func (keys ServerKeys) MarshalJSON() ([]byte, error) {
+	// We already have a copy of the serialised JSON for the keys so we can return that directly.
```
richvdh: I suppose so. We're relying on the fields never being changed?

PR #32, opened 2017-05-02
Add functions for signing and verifying federation HTTP requests

NegativeMjark:


request.go
```diff
@@ -0,0 +1,251 @@
...
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"time"
+)
+
+// A MatrixRequest is a request to send to remote server or a request received
+// from a remote server.
+// Matrix requests are signed by building a JSON object and signing it
+type MatrixRequest struct {
```
Kegsay: `FederationRequest` may be clearer given this has nothing to do with CS API and it could easily be conflated.

request.go
```diff
@@ -0,0 +1,251 @@
...
+	"github.com/matrix-org/util"
+	"golang.org/x/crypto/ed25519"
+	"io"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"time"
+)
+
+// A MatrixRequest is a request to send to remote server or a request received
```
Kegsay: > to send to **a** remote server

request.go
```diff
@@ -0,0 +1,251 @@
...
+		Method      string                       `json:"method"`
+		Origin      string                       `json:"origin"`
+		RequestURI  string                       `json:"uri"`
+		Signatures  map[string]map[string]string `json:"signatures,omitempty"`
+	}
+}
+
+// NewMatrixRequest creates a matrix request. Takes an HTTP method, a
+// destination homeserver and a request path which can have a query string.
+func NewMatrixRequest(method, destination, requestURI string) MatrixRequest {
```
Kegsay: Please give an example as there are a ton of subtle things which are missing from the comments like:
 - Is `method` case-sensitive?
 - Does `destination` need `https://`? Does it accept path segments? Port numbers?
 - Does `requestURI` need a leading `/`? Does it support fragments in addition to query parameters?

request.go
```diff
@@ -0,0 +1,251 @@
...
+	httpReq, err := http.NewRequest(r.fields.Method, urlStr, content)
+	if err != nil {
+		return nil, err
+	}
+
+	if r.fields.Content != nil {
+		httpReq.Header.Set("Content-Type", "application/json")
+	}
+
+	for keyID, sig := range r.fields.Signatures[r.fields.Origin] {
```
Kegsay: Will this segfault if `HTTPRequest()` is called before being signed (since the map will be `nil`?)

request.go
```diff
@@ -0,0 +1,251 @@
...
+		httpReq.Header.Add("Authorization", fmt.Sprintf(
+			"X-Matrix origin=\"%s\",key=\"%s\",sig=\"%s\"", r.fields.Origin, keyID, sig,
```
Kegsay: Do we need to escape any of these values?

request.go
```diff
@@ -0,0 +1,251 @@
...
+// VerifyHTTPRequest extracts and verifies the contents of a net/http.Request.
+// It consumes the body of the request.
+// The JSON content can be accessed using MatrixRequest.Content()
+// Returns an 400 error if there was a problem parsing the request.
+// It authenticates the request using an ed25519 signature using the KeyRing.
+// The origin server can be accessed using MatrixRequest.Origin()
+// Returns a 401 error if there was a problem authenticating the request.
+// HTTP handlers using this should be careful that they only use the parts of
+// the request that have been authenticated: the method, the request path,
+// the query parameters, and the JSON content.
```
Kegsay: So not the headers? Most everything else looks to be taken into account.

request.go
```diff
@@ -0,0 +1,251 @@
...
+
+	return request, util.JSONResponse{Code: 200, JSON: struct{}{}}
+}
+
+// Returns an error if there was a problem reading the content of the request
+func readHTTPRequest(req *http.Request) (*MatrixRequest, error) {
+	var result MatrixRequest
+
+	result.fields.Method = req.Method
+	result.fields.RequestURI = req.URL.RequestURI()
```
Kegsay: Do we need to worry about the way Go escapes/doesn't escape things such that this may not match the thing that was signed?

request.go
```diff
@@ -0,0 +1,251 @@
...
+		} else {
+			result.fields.Signatures[origin][key] = sig
+		}
+	}
+
+	return &result, nil
+}
+
+func parseAuthorizationXMatrix(headerParts []string) (origin, key, sig string) {
+	if len(headerParts) != 2 {
```
Kegsay: Checks feel weirdly placed. The one place where `parseAuthorizationXMatrix` is called we have already checked that `[0]` is `X-Matrix`, but we do the length check here and never use `[0]` again, only `[1]`.

I suggest just passing the header value into this function and remove the length check.

request.go
NegativeMjark: Yep. (The reason for not handling the headers is that reverse proxies sometimes do "things" to them)

request.go
NegativeMjark: No, but only because we rely on the domain and key being safe to include. I guess I should add some checks.

request.go
NegativeMjark: Nope

> A nil map is equivalent to an empty map except that no elements may be added.

https://golang.org/ref/spec#Map_types

request.go
Kegsay: It'd be nice to explicitly call out headers as a thing which is not authenticated, rather than relying on knowing-through-omission.

request.go
NegativeMjark: Done c7dd951

request.go
NegativeMjark: Fixed c7dd951

request.go
NegativeMjark: I've added some more comments and extra checks in c7dd951

Kegsay:
https://github.com/matrix-org/gomatrixserverlib/pull/32#discussion_r114769203 then LGTM

request.go
NegativeMjark: Done 4dd2570

Kegsay:
LGTM!

request.go
NegativeMjark: I've added some checks e4f9b7c

PR #34, opened 2017-05-05
Use a named type for the KeyID

NegativeMjark:


keys.go
```diff
@@ -56,6 +56,9 @@ type OldVerifyKey struct {
 	ExpiredTS Timestamp `json:"expired_ts"`
 }
 
+// A KeyID is the ID of a ed2219 key used by a server.
```
Kegsay: Is this ID arbitrary? Does it have a particular structure like `ed25519:randomtext`? Can `randomtext` contain :hankey: ? ASCII only? Spaces?

Kegsay:
:star: Awesome. Other than single comment, LGTM.

Kegsay:
Wonderful! LGTM.

PR #35, opened 2017-05-05
Add functions for converting between millisecond timestamps and times

NegativeMjark:


Kegsay:
LGTM

Kegsay:
This will probably break parts of Dendrite though which use `OriginServerTS`.

PR #36, opened 2017-05-05
Use a named type for matrix server names

NegativeMjark:


keys.go
```diff
@@ -27,13 +27,18 @@ import (
 	"time"
 )
 
+// A ServerName is the name a matrix homeserver is identified by.
+// It is a DNS name optionally followed by a port.
```
Kegsay: > It is a DNS name

With trailing `.` then? 0:)

Kegsay:
LGTM

PR #37, opened 2017-05-15
Add SyncFormatInvite

Kegsay:
Needed for 'invite_state' of /sync responses.

### DIFF
--- a/inviteevent.go
+++ b/inviteevent.go
@@ -1,0 +1,11 @@
+package gomatrixserverlib
+
+// InviteStateFormattedEvent is an event which is explicitly formatted for
+// the 'invite_state' of a /sync response. It is not a real event.
+type InviteStateFormattedEvent struct {
+	Content  rawJSON `json:"content"`
+	RoomID   string  `json:"room_id,omitempty"`
+	Sender   string  `json:"sender"`
+	StateKey *string `json:"state_key,omitempty"`
+	Type     string  `json:"type"`
+}
--- a/inviteevent.go
+++ b/inviteevent.go
@@ -4,7 +4,6 @@ package gomatrixserverlib
 // the 'invite_state' of a /sync response. It is not a real event.
 type InviteStateFormattedEvent struct {
 	Content  rawJSON `json:"content"`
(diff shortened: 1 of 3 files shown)


NegativeMjark:
As per IRL: I'm not sure this will actually work the way we want because I think the `invite_state` doesn't come from actual room events but instead from partial room state attached by the sending server to the unsigned part of the invite. https://github.com/matrix-org/synapse/blob/v0.20.0/synapse/rest/client/v2_alpha/sync.py#L258-L263

Kegsay:
@NegativeMjark PTAL - I don't actually need this struct for now given I'm no longer doing `invite_state`, so can either merge or close this PR.

inviteevent.go
```diff
@@ -0,0 +1,11 @@
+package gomatrixserverlib
+
+// InviteStateFormattedEvent is an event which is explicitly formatted for
+// the 'invite_state' of a /sync response. It is not a real event.
+type InviteStateFormattedEvent struct {
+	Content  rawJSON `json:"content"`
+	RoomID   string  `json:"room_id,omitempty"`
```
NegativeMjark: Do you need a `RoomID` here?

NegativeMjark:
Of course having told you that it isn't need for calculating the sync response. I can now point out that it would be quite useful for calculating the `unsigned.invite_room_state` key when sending an invite. https://github.com/matrix-org/synapse/blob/v0.20.0/synapse/handlers/message.py#L541-L549

inviteevent.go
Kegsay: I don't know, do we? In `/sync` there is room context (upper keys have the room ID), but over federation I think you'll be missing that or something? IDK.

Kegsay:
Right. I'm going to link to this PR in the docs spreadsheet for implementing `/invite` CS API endpoints so when I implement that I can add in the `invite_state` to the event, sound good to you?

inviteevent.go
Kegsay: Removed.

Kegsay:
Closed for now.

PR #38, opened 2017-05-16
Add a type for matrix federation transactions

NegativeMjark:


transaction.go
```diff
@@ -0,0 +1,71 @@
...
+	PreviousIDs []TransactionID `json:"previous_ids"`
+	// The room events pushed by this transaction.
+	PDUs []Event `json:"pdus"`
+	// The ephemeral events pushed by this transaction.
+	EDUs []EDU `json:"edus"`
+}
+
+// A TransactionID identifies a transaction sent by a matrix server to another
+// matrix server. The ID must be unique amoungst the transactions sent from the
+// origin server to the destination, but doesn't have to be globally unique.
```
Kegsay: Any restrictions? Can I use smiley poo?

transaction.go
```diff
@@ -0,0 +1,71 @@
...
+// A Transaction is used to push data from one matrix server to another matrix
+// server.
+type Transaction struct {
+	// The ID of the transaction.
+	TransactionID TransactionID `json:"transaction_id"`
+	// The server that sent the transaction.
+	Origin ServerName `json:"origin"`
+	// The server that should receive the transaction.
+	Destination ServerName `json:"destination"`
+	// The millisecond posix timestamp on the origin server when the transction
```
Kegsay: > transction

transaction.go
```diff
@@ -0,0 +1,71 @@
...
+	// was created.
+	OriginServerTS Timestamp `json:"origin_server_ts"`
+	// The IDs of the most recent transactions sent by the server.
+	// Mutliple transactions can be sent in parallel so there may be
```
Kegsay: > Mutliple

transaction.go
```diff
@@ -0,0 +1,71 @@
...
+	// The ID of the transaction.
+	TransactionID TransactionID `json:"transaction_id"`
+	// The server that sent the transaction.
+	Origin ServerName `json:"origin"`
+	// The server that should receive the transaction.
+	Destination ServerName `json:"destination"`
+	// The millisecond posix timestamp on the origin server when the transction
+	// was created.
+	OriginServerTS Timestamp `json:"origin_server_ts"`
+	// The IDs of the most recent transactions sent by the server.
```
Kegsay: > sent by the server

Ambiguous. Do you mean "sent by the Destination server" or "sent by the Origin server"? It's not immediately clear as both could be valid (if they are the most recent txns sent by us then it forms a DAG, if they are the most recent txns received then it serves as an ACK).

transaction.go
```diff
@@ -0,0 +1,71 @@
...
+	// Mutliple transactions can be sent in parallel so there may be
+	// more than one previous transaction.
+	PreviousIDs []TransactionID `json:"previous_ids"`
+	// The room events pushed by this transaction.
```
Kegsay: ...what does "pushed" mean in this context? The transaction contains information, and this is it right? I would just say "The XXXX events in this transaction". Ditto `:25`.

transaction.go
```diff
@@ -0,0 +1,71 @@
...
+// origin server to the destination, but doesn't have to be globally unique.
+type TransactionID string
+
+// An EDU is used to transmit ephemeral data such as presence and typing from
+// one matrix server to another.
+type EDU struct {
+	// The type of the EDU, this tells the receiver how to interpret the
+	// contents.
+	EDUType string
+	// The JSON content of the EDU.
```
Kegsay: Ambiguous. Is this akin to the `content` key in say a `m.presence` or `m.typing` event, or is this a layer up and instead represents the entire `m.presence` or `m.typing` event? If the former, please quote "content" to make this a bit clearer.

transaction.go
```diff
@@ -0,0 +1,71 @@
...
+	e.Content, err = json.Marshal(content)
+	return
+}
+
+type eduFields struct {
+	EDUType string  `json:"edu_type"`
+	Content rawJSON `json:"content"`
+}
+
+// MarshalJSON implements json.Marshaller
```
Kegsay: Why? I figure it's for `rawJSON` but we should say so.

transaction.go
NegativeMjark: Fixed 183dc34

transaction.go
NegativeMjark: Commented 183dc34

transaction.go
NegativeMjark: Fixed 183dc34

transaction.go
NegativeMjark: Yes it needs to go in a URL path segment. Yes you can put smiley poo in a URL path segment. But you probably shouldn't.

NegativeMjark:
I've removed the EDUs from transactions entirely since I dunno what the contents of them were as it would have needed me to dive further into the depths of synapse.

Since the reason I'm making this PR is to get message sending/receiving working it's probably best to ignore the EDUs rather adding the code only to have it be wrong or bitrot.

Kegsay:
LGTM!

PR #39, opened 2017-05-16
Add client for the APIs needed to join a room over federation

NegativeMjark:


.travis.yml
```diff
@@ -1,9 +1,10 @@
 language: go
 go:
- - 1.7
+ - 1.8
```
Kegsay: Why?

authedclient.go
```diff
@@ -0,0 +1,140 @@
...
+	"github.com/matrix-org/gomatrix"
+	"golang.org/x/crypto/ed25519"
+	"io/ioutil"
+	"net/http"
+	"net/url"
+)
+
+// An AuthedClient is a matrix federation client that adds
+// "Authorization: X-Matrix" headers to requests that need ed25519 signatures
+type AuthedClient struct {
```
Kegsay: This is explicitly only relevant for federation and yet we don't mention "federation" in the type, but do randomly mention "auth". AFAICT, *every* request made by this client has X-Matrix headers added, so mentioning a random impl detail in the type feels strange. Can't we just call this `FederationClient`?

federationtypes.go
```diff
@@ -0,0 +1,97 @@
+package gomatrixserverlib
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// A SendRequest is the content of a request to PUT /_matrix/federation/v1/send/{txnID}/
+type SendRequest Transaction
```
Kegsay: I don't like this for 2 reasons:
 - We usually do this so we can have the compiler help us out. The one place where this is used takes an `interface{}` so there is no benefit to typing this.
 - Casting to this type make it look like it has side-effects:
    ```go
   r := SendRequest(txn)
    ```

I would stick with `Transaction` and remove this type.

authedclient.go
```diff
@@ -0,0 +1,140 @@
...
+	req := NewFederationRequest("PUT", t.Destination, path)
+	if err = req.SetContent(SendRequest(t)); err != nil {
+		return
+	}
+	err = ac.doRequest(req, &res)
+	return
+}
+
+// MakeJoin makes a join m.room.member event for a room on a remote matrix server.
+// This is used to join a room the local server isn't a member of.
```
Kegsay: Now if this were the CS API I'd be linking to the spec... but in absence of that, I'm going to need you to go into a bit more detail here.
- Under what circumstances will this fail? Should we retry if it does? Backoff?
- Must this be called prior to `send_join`?
- Can the `userID` be anything? Does it have to be a user which we (the originator) control (as given by the domain)?

Kegsay: - - If I call `MakeJoin` multiple times, do I get multiple events back? Or do I get the same event every time?

authedclient.go
```diff
@@ -0,0 +1,140 @@
...
+	path := "/_matrix/federation/v1/make_join/" +
+		url.PathEscape(roomID) + "/" +
+		url.PathEscape(userID)
+	req := NewFederationRequest("GET", s, path)
+	err = ac.doRequest(req, &res)
+	return
+}
+
+// SendJoin sends a join m.room.member event via a remote matrix server.
+// This is used to join a room the local server isn't a member of.
```
Kegsay: Now if this were the CS API I'd be linking to the spec... but in absence of that, I'm going to need you to go into a bit more detail here.
- Under what circumstances will this function return an error? Any specific error codes?
- What is `event`? Must this be the same thing returned by `MakeJoin`? Can we make up our own one? Do we need to modify the event in any way?

authedclient.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return err
+	}
+
+	return json.Unmarshal(contents, resBody)
+}
+
+// SendTransaction sends a transaction
+func (ac *AuthedClient) SendTransaction(t Transaction) (res SendResponse, err error) {
+	path := "/_matrix/federation/v1/send/" + string(t.TransactionID) + "/"
```
Kegsay: Trailing slash?

authedclient.go
```diff
@@ -0,0 +1,140 @@
...
+	req := NewFederationRequest("PUT", s, path)
+	if err = req.SetContent(event); err != nil {
+		return
+	}
+	err = ac.doRequest(req, &res)
+	return
+}
+
+// LookupState retrieves the room state for a room at an event from a
+// remote matrix server as full matrix events.
```
Kegsay: Now if this were the CS API I'd be linking to the spec... but in absence of that, I'm going to need you to go into a bit more detail here.
 - When does this fail?
 - Under what failures is it safe to retry (transient failures)

authedclient.go
```diff
@@ -0,0 +1,140 @@
...
+		url.PathEscape(roomID) +
+		"/?event_id=" +
+		url.QueryEscape(eventID)
+	req := NewFederationRequest("GET", s, path)
+	err = ac.doRequest(req, &res)
+	return
+}
+
+// LookupStateIDs retrieves the room state for a room at an event from a
+// remote matrix server as lists of matrix event IDs.
```
Kegsay: Now if this were the CS API I'd be linking to the spec... but in absence of that, I'm going to need you to go into a bit more detail here.
 - When does this fail?
 - Under what failures is it safe to retry (transient failures)

authedclient.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	if err != nil {
+		return err
+	}
+
+	return json.Unmarshal(contents, resBody)
+}
+
+// SendTransaction sends a transaction
```
Kegsay: Now if this were the CS API I'd be linking to the spec... but in absence of that, I'm going to need you to go into a bit more detail here.
 - When does this fail?
 - Under what failures is it safe to retry (transient failures)
 - Presumably the transaction must be immutable by this point: we cannot add more PDUs between retries.
 - Do we update `OriginServerTS` between retries?

federationtypes.go
```diff
@@ -0,0 +1,97 @@
...
+
+// A SendResponse is the content of a response to PUT /_matrix/federation/v1/send/{txnID}/
+type SendResponse struct {
+	// Map of string room event ID to the result of processing that event.
```
Kegsay: > string room event ID

Er... what? I can tell it's a string already, but is this a room ID or event ID as the key? Or maybe it's a concatenation of both?

federationtypes.go
Kegsay: Alternatively, if we want the warm fuzzies of typing requests/responses, I would advocate using `ReqXXX` and `RespXXXX` to make it clearer that:
 - They don't look like callable functions at call-sites.
 - They sort nicely on godoc.

federationtypes.go
```diff
@@ -0,0 +1,97 @@
...
+type StateResponse struct {
+	// A list of events giving the state of the room before the request event.
+	StateEvents []Event `json:"pdus"`
+	// A list of events needed to authenticate the state events.
+	AuthEvents []Event `json:"auth_chain"`
+}
+
+// A MakeJoinResponse is the content of a response to GET /_matrix/federation/v1/make_join/{roomID}/{userID}
+type MakeJoinResponse struct {
+	JoinEvent Event `json:"event"`
```
Kegsay: I'd like to know more about what this is. I assume it is an `m.room.member` event for the given user in the given room, but presumably this doesn't "activate" until `/send_join` is called?

federationtypes.go
```diff
@@ -0,0 +1,97 @@
...
+// A SendJoinRequest is the content of a request to PUT /_matrix/federation/v1/send_join/{roomID}/{eventID}
+type SendJoinRequest Event
+
+// A SendJoinResponse is the content of a response to PUT /_matrix/federation/v1/send_join/{roomID}/{eventID}
+type SendJoinResponse StateResponse
+
+// MarshalJSON implements json.Marshaller
+func (r SendJoinResponse) MarshalJSON() ([]byte, error) {
+	// SendJoinResponses contain the same data as a StateResponse but are
+	// formatted slightly differently on the wire:
```
Kegsay: Excellent explanation :star:

federationtypes.go
```diff
@@ -0,0 +1,97 @@
...
+	// (This protocol oddity is the result of a typo in the synapse matrix
+	//  server, and is preserved to maintain compatibility.)
+
+	return json.Marshal([]interface{}{200, sendJoinResponseFields{
+		r.StateEvents, r.AuthEvents,
+	}})
+}
+
+// UnmarshalJSON implements json.Unmarshaller
+func (r *SendJoinResponse) UnmarshalJSON(data []byte) error {
```
Kegsay: A tiny unit test to check your JSON juggling is working would be appreciated.

.travis.yml
NegativeMjark: Because net/url.PathEncode is a go1.8 thing

authedclient.go
NegativeMjark: Done 10b2dfb

federationtypes.go
NegativeMjark: Done 5fad723

authedclient.go
NegativeMjark: (And I've renamed the file 5a80be1)

authedclient.go
NegativeMjark: Yes, it think it's necessary to match the regexp at https://github.com/matrix-org/synapse/blob/master/synapse/federation/transport/server.py#L207

Kegsay: Wow. Okay.

Kegsay:
Tiny test please as per https://github.com/matrix-org/gomatrixserverlib/pull/39#discussion_r116949204 then LGTM

Kegsay:
LGTM

PR #40, opened 2017-05-18
Add the "/query/directory" federation API to the client

NegativeMjark:


federationclient.go
```diff
@@ -147,3 +147,14 @@ func (ac *FederationClient) LookupStateIDs(s ServerName, roomID, eventID string)
 	err = ac.doRequest(req, &res)
 	return
 }
+
+// LookupRoomAlias looks up a room alias hosted on the remote server.
+// The domain part of the roomAlias must match the name of the server it is
+// being looked up on.
```
Kegsay: If the room alias does not exist on this server, what happens?
 - An error is returned? What does that look like?
 - A  `RespDirectory` is returned which has `len(servers) == 0`?

Kegsay:
Otherwise LGTM!

federationclient.go
NegativeMjark: Commented b7bcebb

Kegsay:
LGTM

PR #42, opened 2017-05-23
Enforce event size limits

NegativeMjark:


event.go
```diff
@@ -303,6 +313,117 @@ func (e Event) StateKeyEquals(stateKey string) bool {
...
+	if err != nil {
+		return err
+	}
+
+	senderDomain, err := checkID(e.fields.Sender, "user", '@')
+	if err != nil {
+		return err
+	}
+
+	eventDomain, err := checkID(e.fields.EventID, "event", '$')
```
Kegsay: Hmm, should we be doing this? http://matrix.org/docs/spec/intro.html#room-ids-and-event-ids

> The domain is used only for namespacing to avoid the risk of clashes of identifiers between different homeservers.

But yet we assert here that the event is in this format?

event.go
```diff
@@ -303,6 +313,117 @@ func (e Event) StateKeyEquals(stateKey string) bool {
...
+	if e.fields.Origin != ServerName(eventDomain) {
+		return fmt.Errorf(
+			"gomatrixserverlib: event ID domain doesn't match origin: %q != %q",
+			eventDomain, e.fields.Origin,
+		)
+	}
+
+	if eventDomain != senderDomain {
+		// For the most part all events should be sent by a user on the
+		// orignating server
```
Kegsay: > orignating

event.go
```diff
@@ -303,6 +313,117 @@ func (e Event) StateKeyEquals(stateKey string) bool {
 	return *e.fields.StateKey == stateKey
 }
 
+const (
+	// The event ID, room ID sender, event type and state key fields cannot be
```
Kegsay: > room ID sender

Kegsay:
Otherwise LGTM

event.go
Kegsay: I meant to tag https://github.com/matrix-org/gomatrixserverlib/pull/42/files#diff-9fb85d1ee9f1b0a59804bec41e1491feR368

Kegsay:
LGTM

PR #43, opened 2017-05-23
Add methods for checking if federation responses are valid

NegativeMjark:


federationtypes.go
```diff
@@ -104,3 +104,106 @@ type RespDirectory struct {
 	// before it finds one that it can use to join the room.
 	Servers []ServerName `json:"servers"`
 }
+
+// Check that a response to /state is valid.
+func (r RespState) Check(keyRing KeyRing) error {
```
Kegsay: Can you move this function to be near to where `RespState` is defined please?

eventcrypto.go
```diff
@@ -184,3 +184,55 @@ func verifyEventSignature(signingName string, keyID KeyID, publicKey ed25519.Pub
...
+// signatures from the server that sent it.
+func VerifyEventSignatures(events []Event, keyRing KeyRing) error {
+	var toVerify []VerifyJSONRequest
+	for _, event := range events {
+		redactedJSON, err := redactEvent(event.eventJSON)
+		if err != nil {
+			return err
+		}
+		v := VerifyJSONRequest{
+			Message:    redactedJSON,
```
Kegsay: Naming and docs: It's not clear that `VerifyJSONRequest` actually wants the redacted JSON and not the full event JSON. I would kill two birds with one stone and `s/Message/RedactedJSON/` assuming `VerifyJSONRequest` is only ever used for `Event` structs.

federationtypes.go
NegativeMjark: Done 9df1e37

eventcrypto.go
```diff
@@ -184,3 +184,55 @@ func verifyEventSignature(signingName string, keyID KeyID, publicKey ed25519.Pub
...
+			AtTS:       event.OriginServerTS(),
+			ServerName: event.Origin(),
+		}
+		toVerify = append(toVerify, v)
+
+		// "m.room.member" invite events are signed by both the server sending
+		// the invite and the server the invite is for.
+		if event.Type() == "m.room.member" && event.StateKey() != nil {
```
Kegsay: I'm getting increasingly anxious about using hard-coded strings over constants as we continue to grow this library. It would be nice if you could declare some consts so we cannot possibly typo `m.room.memmber` - I don't expect it to be done in this PR.

eventcrypto.go
```diff
@@ -184,3 +184,55 @@ func verifyEventSignature(signingName string, keyID KeyID, publicKey ed25519.Pub
...
+
+	results, err := keyRing.VerifyJSONs(toVerify)
+	if err != nil {
+		return err
+	}
+
+	// Check that all the event JSON was correctly signed.
+	for _, result := range results {
+		if result.Result != nil {
+			return result.Result
```
Kegsay: Naming: I don't like that `VerifyJSONResult.Result` (which is of type `error`) is called `Result`. Why isn't it called something like `Err` or `Error` given its type? It's non-obvious at call-sites like this what is going on.

I will tolerate the return parameters for `KeyRing.VerifyJSONs` but it would be nice if we could work out a better way than effectively returning `[]error, error`. I feel it is important to keep the distinction, but it's annoying to expect callers to do the loop to check for errors in addition to an error return when we just do the same thing (return any errors). I don't have any better proposals though.

Kegsay: Maybe abuse types and if callers care about the difference between the errors they can type-cast? That way it looks like:
```go
errs := keyRing.VerifyJSONs(toVerify)
for _, err := range errs {
  return err
}
```

which I think is clearer.

eventcrypto.go
```diff
@@ -184,3 +184,55 @@ func verifyEventSignature(signingName string, keyID KeyID, publicKey ed25519.Pub
...
+		}
+	}
+
+	// Everything was okay.
```
Kegsay: I can guess that from the `return nil`.

federationtypes.go
```diff
@@ -104,3 +104,106 @@ type RespDirectory struct {
...
+		if event.StateKey() == nil {
+			return fmt.Errorf("gomatrixserverlib: event %q does not have a state key", event.EventID())
+		}
+		if stateTuples[StateKeyTuple{event.Type(), *event.StateKey()}] {
+			return fmt.Errorf(
+				"gomatrixserverlib: duplicate state key tuple (%q, %q)",
+				event.Type(), *event.StateKey(),
+			)
+		}
+		allEvents = append(allEvents, event)
```
Kegsay: Aren't you missing a write to `stateTuples` so the dupe checks actually work?

federationtypes.go
```diff
@@ -104,3 +104,106 @@ type RespDirectory struct {
...
+	}
+
+	// Check whether the events are allowed by the auth rules.
+	for _, event := range allEvents {
+		if err := checkAllowedByAuthEvents(event, eventsByID); err != nil {
+			return err
+		}
+	}
+
+	// The checks pass.
```
Kegsay: I can tell that from `return nil`. It's a common enough pattern to have `func CheckXXX() error` that it doesn't warrant explanation that returning no error means everything was fine.

federationtypes.go
```diff
@@ -104,3 +104,106 @@ type RespDirectory struct {
...
+	return nil
+}
+
+// Check that a reponse to /send_join is valid.
+func (r RespSendJoin) Check(keyRing KeyRing, joinEvent Event) error {
+	// First check that the state is valid.
+	if err := RespState(r).Check(keyRing); err != nil {
```
Kegsay: Wait what? `RespSendJoin` can be directly type-cast to `RespState`?! Explain please.

federationtypes.go
NegativeMjark: Fixed 3079aa0

eventcrypto.go
NegativeMjark: Fixed 07f2393

federationtypes.go
NegativeMjark: Fixed 07f2393

eventcrypto.go
NegativeMjark: >  assuming VerifyJSONRequest is only ever used for Event

It gets used elsewhere I'm afraid.

eventcrypto.go
NegativeMjark: The main complication is that at somepoint we might want to implement the filtering logic described by https://github.com/matrix-org/synapse/blob/v0.21.0/synapse/federation/federation_base.py#L41-L44

So it might be quite useful to know which PDUs failed and succeeded. I'm struggling to think of a nice API shape for doing that though.

NegativeMjark: > Maybe abuse types and if callers care about the difference between the errors they can type-cast?

Yeah. What do you think about doing this for error handling in general in order to make it easier for dendrite to interpret the errors when translating them to HTTP status codes?

Should we be baking the raw HTTP status codes and responses deep into the stack like we do in synapse?
Or should we be defining a set of types in dendrite that we then translate into HTTP status codes/responses inside dendrite?

eventcrypto.go
NegativeMjark: Sure, fine. Will do straight after this PR. :)

eventcrypto.go
Kegsay: Moving this discussion to https://docs.google.com/document/d/1VUkAafJgk-e0S4zjBah6t9Oriw-WUgM-76Vl7iDyeIQ/edit?usp=sharing

Kegsay:
LGTM

PR #44, opened 2017-05-24
Use named constants for event types

NegativeMjark:


eventauth.go
```diff
@@ -23,11 +23,19 @@ import (
...
+	invite                 = "invite"
+	public                 = "public"
+	mRoomCreate            = "m.room.create"
+	mRoomJoinRules         = "m.room.join_rules"
+	mRoomPowerLevels       = "m.room.power_levels"
+	mRoomMember            = "m.room.member"
+	mRoomThirdPartyInvite  = "m.room.third_party_invite"
+	mRoomAliases           = "m.room.aliases"
+	mRoomHistoryVisibility = "m.room.history_visibility"
+	mRoomRedaction         = "m.room.redaction"
```
Kegsay: Can these be public please or else I'll be repeating this in Dendrite code.

Kegsay:
Otherwise LGTM.

eventauth.go
NegativeMjark: Done 258b2b4

Kegsay:
LGTM

PR #45, opened 2017-05-24
Add method for listing the events in a /state response

NegativeMjark:
Returns the events in an order where every event comes after its
auth_events. This is necessary to insert the events into the roomserver
in dendrite.

### DIFF
--- a/federationtypes.go
+++ b/federationtypes.go
@@ -52,8 +52,15 @@ func (r RespState) Events() ([]Event, error) {
 	var result []Event
 	for _, event := range eventsByID {
 		if visited[event] {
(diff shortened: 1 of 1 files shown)


federationtypes.go
```diff
@@ -34,6 +34,77 @@ type RespState struct {
 	AuthEvents []Event `json:"auth_chain"`
 }
 
+// Events combines the auth events and the state events and returns
+// them in an order where every event comes after its auth events.
```
Kegsay: Does this mean there will be dupes or do we filter those out?

Kegsay:
LGTM - Well explained and clear implementation.

PR #46, opened 2017-06-01
client: Add CreateMediaDownloadRequest API

superdump:


client.go
```diff
@@ -220,3 +220,29 @@ func (fc *Client) LookupServerKeys(
...
+
+	remoteReqAddr := "https://" + dnsResult.Addrs[0] + "/_matrix/media/v1/download/" + string(matrixServer) + "/" + mediaID
+	remoteReq, err := http.NewRequest("GET", remoteReqAddr, nil)
+	if err != nil {
+		return nil, err
+	}
+
+	remoteReq.Header.Set("Host", string(matrixServer))
+
+	resp, err := fc.client.Do(remoteReq)
```
NegativeMjark: The http client in `fc.client` has been modified to support a "matrix://" URL scheme that handles the server lookup so all of the above code can be replaced with 

```go
requestURL := "matrix://" + string(matrixServer) + "/_matrix/media/v1/download/" + string(matrixServer) + "/" + mediaID
resp, err := fc.client.GET(requestURL)
```

client.go
```diff
@@ -220,3 +220,29 @@ func (fc *Client) LookupServerKeys(
...
+	if err != nil {
+		return nil, fmt.Errorf("Failed to execute request for remote file")
```
NegativeMjark: I'd just return the `err` here.

client.go
superdump: Done.

client.go
superdump: Done.

Issue #47, opened 2017-06-09
Sigining.go


madguy02:
I wanted to know how "var object map[string]*json.RawMessage" is behaving,and also the nature of the Rawmessage?

I am actually trying to convert this behavior into JS.So,a little explanation of "func SignJson" is much appreciated

NegativeMjark:
Hi @madguy02.

The behaviour of SignJSON is part of the matrix specification and is documented at https://matrix.org/docs/spec/server_server/unstable.html#signing-json

json.RawMessage is part of the go standard library and is documented at https://golang.org/pkg/encoding/json/#RawMessage

The matrix-js-sdk already has a partial implementation of JSON signing https://github.com/matrix-org/matrix-js-sdk/blob/0f30d21fa270f582d0023214ccca8259d2c6e733/src/crypto/index.js#L1115-L1126

I've added a few comments to the go version to help explain it in #48.

If it helps there is a python implementation at https://github.com/matrix-org/python-signedjson/blob/master/signedjson/sign.py#L26-L53

madguy02:
hii @NegativeMjark thanks for your response

PR #48, opened 2017-06-09
Add comments to explain more of SignJSON and VerifyJSON

NegativeMjark:
This might help with #47


signing.go
```diff
@@ -48,32 +55,35 @@ func SignJSON(signingName string, keyID KeyID, privateKey ed25519.PrivateKey, me
...
 		return nil, err
 	}
-
 	canonical, err := CanonicalJSON(unsorted)
 	if err != nil {
 		return nil, err
 	}
 
+	// Sign the canonical JSON with the ed25519 key.
 	signature := Base64String(ed25519.Sign(privateKey, canonical))
```
NegativeMjark: I'm usually against adding comments with so much redundant information, but I think it's useful here to break up the function into sections.

Kegsay:
LGTM

signing.go
Kegsay: SGTM

PR #54, opened 2017-07-06
Add some utility methods to Event that are useful in dendrite

NegativeMjark:
I was getting annoyed by all the "TODO: this should be in gomatrixserverlib" markers in dendrite.

### DIFF
--- a/event.go
+++ b/event.go
@@ -500,9 +500,9 @@ func (e Event) AuthEvents() []EventReference {
 
 // AuthEventIDs returns the event IDs of the events needed to auth the event.
 func (e Event) AuthEventIDs() []string {
-	result := make([]string, len(e.fields.PrevEvents))
-	for i := range e.fields.PrevEvents {
-		result[i] = e.fields.PrevEvents[i].EventID
+	result := make([]string, len(e.fields.AuthEvents))
+	for i := range e.fields.AuthEvents {
+		result[i] = e.fields.AuthEvents[i].EventID
 	}
 	return result
 }
--- a/event.go
+++ b/event.go
@@ -482,15 +482,15 @@ func (e Event) PrevEventIDs() []string {
 // is an "m.room.member" event.
 // Returns an error if the event is not a m.room.member event or if the content
 // is not valid m.room.member content.
-func (e Event) Membership() (*string, error) {
+func (e Event) Membership() (string, error) {
 	if e.fields.Type != MRoomMember {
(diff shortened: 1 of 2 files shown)


event.go
```diff
@@ -467,11 +469,44 @@ func (e Event) PrevEvents() []EventReference {
...
+		result[i] = e.fields.PrevEvents[i].EventID
+	}
+	return result
+}
+
+// Membership returns the value of the content.membership field if this event
+// is an "m.room.member" event.
+// Returns an error if the event is not a m.room.member event or if the content
+// is not valid m.room.member content.
+func (e Event) Membership() (*string, error) {
```
Kegsay: I'd prefer this to be `(string, error)` on the basis that `""` is not a valid membership state, and constantly dereffing is annoying.

event.go
```diff
@@ -467,11 +469,44 @@ func (e Event) PrevEvents() []EventReference {
...
 // AuthEvents returns references to the events needed to auth the event.
 func (e Event) AuthEvents() []EventReference {
 	return e.fields.AuthEvents
 }
 
+// AuthEventIDs returns the event IDs of the events needed to auth the event.
+func (e Event) AuthEventIDs() []string {
+	result := make([]string, len(e.fields.PrevEvents))
+	for i := range e.fields.PrevEvents {
+		result[i] = e.fields.PrevEvents[i].EventID
```
Kegsay: This is the same as `PrevEventIDs()`?

event.go
```diff
@@ -534,3 +569,19 @@ func (er EventReference) MarshalJSON() ([]byte, error) {
 
 	return json.Marshal(&tuple)
 }
+
+// SplitID splits a matrix ID into a local part and a server name.
+func SplitID(sigil byte, id string) (local string, domain ServerName, err error) {
```
Kegsay: Examples as tests please (they then show up as an example on godoc and are executed as tests, so win-win!).

They look like https://godoc.org/github.com/matrix-org/gomatrix#EncodeUserLocalpart and are implemented like https://github.com/matrix-org/gomatrix/blob/master/userids_examples_test.go#L5

Kegsay:
Otherwise LGTM - I like this plan - Add TODOs until you snap :)

Kegsay:
It'd be nice if the examples included tests for domains with ports.

Kegsay:
LGTM

PR #55, opened 2017-07-24
Add method for setting the unsigned field on an event

NegativeMjark:


event.go
```diff
@@ -252,10 +252,41 @@ func (e Event) Redact() Event {
 		// This is unreachable for events created with EventBuilder.Build or NewEventFromUntrustedJSON
 		panic(fmt.Errorf("gomatrixserverlib: invalid event %v", err))
 	}
-	return Event{
+	result := Event{
 		redacted:  true,
 		eventJSON: eventJSON,
 	}
+	if err = json.Unmarshal(eventJSON, &result.fields); err != nil {
```
Kegsay: Presumably this is just a whoopsie that we forgot to do..?

Kegsay:
Otherwise LGTM

event.go
NegativeMjark: Yep

PR #58, opened 2017-09-06
Support for third-party invites

babolivier:
This PR adds support for `m.room.member` events of `invite` membership derived from `m.room.third_party_invite` events.

### DIFF
--- a/base64.go
+++ b/base64.go
@@ -18,6 +18,7 @@ package gomatrixserverlib
 import (
 	"encoding/base64"
 	"encoding/json"
+	"strings"
 )
 
 // A Base64String is a string of bytes that are base64 encoded when used in JSON.
@@ -43,6 +44,12 @@ func (b64 *Base64String) UnmarshalJSON(raw []byte) (err error) {
 	if err = json.Unmarshal(raw, &str); err != nil {
 		return
 	}
-	*b64, err = base64.RawStdEncoding.DecodeString(str)
+	// We must check whether the string was encoded in a URL-safe way in order
+	// to use the appropriate encoding.
+	if strings.ContainsAny(str, "-_") {
+		*b64, err = base64.RawURLEncoding.DecodeString(str)
+	} else {
+		*b64, err = base64.RawStdEncoding.DecodeString(str)
+	}
 	return
 }
--- a/eventcontent.go
+++ b/eventcontent.go
@@ -108,7 +108,18 @@ type memberContent struct {
 	// We use the membership key in order to check if the user is in the room.
 	Membership string `json:"membership"`
 	// We use the third_party_invite key to special case thirdparty invites.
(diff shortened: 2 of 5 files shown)


base64.go
```diff
@@ -43,6 +44,12 @@ func (b64 *Base64String) UnmarshalJSON(raw []byte) (err error) {
 	if err = json.Unmarshal(raw, &str); err != nil {
 		return
 	}
-	*b64, err = base64.RawStdEncoding.DecodeString(str)
+	// We must check whether the string was encoded in a URL-safe way in order
+	// to use the appropriate encoding.
+	if strings.ContainsAny(str, "-_") {
+		*b64, err = base64.RawURLEncoding.DecodeString(str)
```
NegativeMjark: Won't this create problems when we encode the data? Would it make more sense to add a `URLBase64String` instead?

eventcontent.go
```diff
@@ -137,6 +148,37 @@ func newMemberContentFromEvent(event Event) (c memberContent, err error) {
...
+	return newThirdPartyInviteContentFromEvent(*thirdPartyInviteEvent)
+}
+
+func newThirdPartyInviteContentFromEvent(event Event) (t thirdPartyInviteContent, err error) {
+	if err = json.Unmarshal(event.Content(), &t); err != nil {
+		err = errorf("unparsable third party invite event content: %s", err.Error())
+		return
+	}
+	return
+}
```
NegativeMjark: The `newThirdPartyInviteContentFromEvent` function doesn't seem to be doing much here, would it make sense to merge it into the `newThirdPartyInviteContentFromAuthEvents` function above?

base64.go
babolivier: > Won't this create problems when we encode the data?

I don't think so, since the characters used to detect an URL-safe base64 are specific to this encoding and are assured to not appear in a standard base64-encoded string.

> Would it make more sense to add a `URLBase64String` instead?

Not sure, it would mean we must check the encoding each time we want to decode a base64 string provided by a third party and I'm not sure it would be really relevant.

NegativeMjark: If the spec is requiring URL-safe base64 then shouldn't we make sure that we encode using URL-safe base64 as well as decoding it?

babolivier: The spec doesn't actually mention whether the encoding has to be URL-safe (and even [uses the non-URL-safe encoding](https://matrix.org/speculator/spec/HEAD/appendices.html#unpadded-base64)), so not sure we need URL-safe encoding.
Decoding is required as sydent encodes its keys with the URL-safe encoding, but since it isn't mentioned int the spec and it doesn't change the structure of the endpoint I understand it as a liberty of the implementation.

eventcontent.go
babolivier: Done https://github.com/matrix-org/gomatrixserverlib/pull/58/commits/d05d616c70ca8d2e6daac1b487f710ebc2dbf090

PR #61, opened 2017-09-11
3PID invite exchange over federation

babolivier:
Function for 3PID invite exchange over federation + bugfix.


NegativeMjark:
LGTM

PR #66, opened 2017-09-26
Log outbound requests

richvdh:
Currently it's unclear what we're waiting for when processing requests which
trigger outbound federation requests. Let's add some logging for them.

### DIFF
--- a/client.go
+++ b/client.go
@@ -26,6 +26,8 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+
+	"github.com/matrix-org/util"
 )
 
 // A Client makes request to the federation listeners of matrix
@@ -120,7 +122,7 @@ func (fc *Client) LookupUserInfo(
 	}
 
 	var response *http.Response
-	response, err = fc.client.Do(req.WithContext(ctx))
+	response, err = fc.doHttpRequest(ctx, req)
 	if response != nil {
 		defer response.Body.Close() // nolint: errcheck
 	}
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -244,10 +246,23 @@ func (fc *Client) CreateMediaDownloadRequest(
 	if err != nil {
 		return nil, err
 	}
+
+	return fc.doHTTPRequest(ctx, req)
+}
+
+func (fc *Client) doHTTPRequest(ctx context.Context, req *http.Request) (*http.Response, error) {
+	reqID := util.RandomString(12)
+	logger := util.GetLogger(ctx).WithField("out.req.ID", reqID)
```
NegativeMjark: Would it make sense to set a ```WithField("server", r.fields.Destination)``` here?

federationclient.go
```diff
@@ -42,16 +46,21 @@ func (ac *FederationClient) doRequest(ctx context.Context, r FederationRequest,
 		return err
 	}
 
+	logger.Infof("Outgoing request %s %s", req.Method, req.URL)
```
NegativeMjark: Why does one log line say "Outgoing request" and the other say "Sending request"?

richvdh: Because I changed one to "Outgoing" and forgot to change the other.

client.go
richvdh: yeah I guess so. It's fairly clear from the URI, but I guess it's useful to be able to grep for the server field.

PR #67, opened 2017-09-26
Handle events where the sender and origin do not match

richvdh:
It turns out that synapse had a bug for a while (at the end of 2015) where it would send out `join` events which used the event_id (and origin) from the helping server rather than the joining server. We need to accept these events so that we can join rooms which include such events.

### DIFF
--- a/eventcrypto_test.go
+++ b/eventcrypto_test.go
@@ -17,7 +17,9 @@ package gomatrixserverlib
 
 import (
 	"bytes"
+	"context"
 	"encoding/base64"
+	"encoding/json"
 	"testing"
 
 	"golang.org/x/crypto/ed25519"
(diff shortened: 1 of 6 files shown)


keyring.go
```diff
@@ -69,12 +69,18 @@ type VerifyJSONResult struct {
 	Error error
 }
 
-// VerifyJSONs performs bulk JSON signature verification for a list of VerifyJSONRequests.
-// Returns a list of VerifyJSONResults with the same length and order as the request list.
-// The caller should check the Result field for each entry to see if it was valid.
-// Returns an error if there was a problem talking to the database or one of the other methods
-// of fetching the public keys.
-func (k *KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest) ([]VerifyJSONResult, error) { // nolint: gocyclo
+// A JSONVerifier is an object which can the signatures of JSON messages.
```
NegativeMjark: "which can the signatures"

eventcrypto_test.go
```diff
@@ -258,3 +261,127 @@ func TestSignEventTestVectors(t *testing.T) {
...
+		t.Fatal(err)
+	}
+	event.eventJSON = eventJSON
+
+	events := []Event{event}
+	if err := VerifyEventSignatures(context.Background(), events, &verifier); err != nil {
+		t.Fatal(err)
+	}
+
+	// should be two call to the verifier
```
NegativeMjark: Should this be "There should be two verification requests".
(I find the use of "call" confusing because VerifyJSONs is only called once)

PR #70, opened 2017-11-09
Set a 30s timeout on client requests.

rxl881:
Scalar is not currently specifying HTTP timeouts. As mentioned here - https://medium.com/@nate510/don-t-use-go-s-default-http-client-4804cb19f779 - Go doesn't specify sensible defaults and we need to make sure that we do so, especially in production.

This was specifically brought to light via calls to scalar/api/register, which were timing out and filling up the apache front-end. This endpoint is designed to make outbound HTTP requests to provision a user on a HS. If the the destination HS is poorly configured or not set up for federation, it might fail such that it doesn't timeout.

See - matrix-org/scalar#260 for further details.

### DIFF
--- a/client.go
+++ b/client.go
@@ -48,7 +48,9 @@ type UserInfo struct {
 
 // NewClient makes a new Client
 func NewClient() *Client {
-	return &Client{client: http.Client{Transport: newFederationTripper()}}
+	return &Client{client: http.Client{
+		Transport: newFederationTripper(),
+		Timeout:   requestTimeout}}
 }
 
 type federationTripper struct {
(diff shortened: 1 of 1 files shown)


rxl881:
See - https://github.com/matrix-org/scalar/issues/260 for details.

richvdh:
> See - matrix-org/scalar#260 for further details.

@rxl881: scalar is a private repo whereas this one isn't...

For anyone who can't see the original issue: the summary is that requests to the /register endpoint on scalar were taking ages and filling up slots on our apache; the reason for that turned out to be that scalar was waiting for an unresponsive homeserver to respond.

rxl881:
@richvdh - As it was a really quick fix, I've added the ability to create a client with a custom timeout (assuming that's what you were after, rather than per-request). Can you please take another look?

richvdh:
I was really thinking it needed to happen per-request to be honest. Still, this is fine for now either way.

rxl881:
OK, I'll leave it like this for the time being. Cheers.

PR #71, opened 2017-11-13
Make KeyFetcher interface return keys

richvdh:
This PR reworks the KeyFetcher interface so that, for each key requested, it returns that key in the form of a PublicKeyLookupRequest, rather than the raw JSON which came from the /keys/v1/query response. This will give us a better basis on which to implement fetching keys from
perspectives servers.

The corollary is that we also need to update KeyDatabase to work at the higher level (of keys instead of JSON).

(We'll need to update the application level to handle the changes in interfaces).

### DIFF
--- a/base64.go
+++ b/base64.go
@@ -26,13 +26,31 @@ import (
 // When the bytes are unmarshalled from JSON they are decoded from base64.
 type Base64String []byte
 
+// Encode encodes the bytes as base64
+func (b64 Base64String) Encode() string {
+	return base64.RawStdEncoding.EncodeToString(b64)
+}
+
+// Decode decodes the given input into this Base64String
+func (b64 *Base64String) Decode(str string) error {
+	// We must check whether the string was encoded in a URL-safe way in order
+	// to use the appropriate encoding.
+	var err error
+	if strings.ContainsAny(str, "-_") {
+		*b64, err = base64.RawURLEncoding.DecodeString(str)
+	} else {
+		*b64, err = base64.RawStdEncoding.DecodeString(str)
+	}
+	return err
+}
+
 // MarshalJSON encodes the bytes as base64 and then encodes the base64 as a JSON string.
 // This takes a value receiver so that maps and slices of Base64String encode correctly.
 func (b64 Base64String) MarshalJSON() ([]byte, error) {
 	// This could be made more efficient by using base64.RawStdEncoding.Encode
 	// to write the base64 directly to the JSON. We don't need to JSON escape
 	// any of the characters used in base64.
-	return json.Marshal(base64.RawStdEncoding.EncodeToString(b64))
+	return json.Marshal(b64.Encode())
 }
 
 // UnmarshalJSON decodes a JSON string and then decodes the resulting base64.
@@ -44,12 +62,6 @@ func (b64 *Base64String) UnmarshalJSON(raw []byte) (err error) {
 	if err = json.Unmarshal(raw, &str); err != nil {
 		return
 	}
-	// We must check whether the string was encoded in a URL-safe way in order
-	// to use the appropriate encoding.
-	if strings.ContainsAny(str, "-_") {
-		*b64, err = base64.RawURLEncoding.DecodeString(str)
-	} else {
-		*b64, err = base64.RawStdEncoding.DecodeString(str)
-	}
+	err = b64.Decode(str)
 	return
 }
--- a/keyring.go
+++ b/keyring.go
@@ -123,6 +123,8 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest) 
 	k.checkUsingKeys(requests, results, keyIDs, keysFromDatabase)
 
 	for i := range k.KeyFetchers {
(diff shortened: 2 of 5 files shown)


base64.go
```diff
@@ -26,13 +26,31 @@ import (
...
+func (b64 Base64String) Encode() string {
+	return base64.RawStdEncoding.EncodeToString(b64)
+}
+
+// Decode decodes the given input into this Base64String
+func (b64 *Base64String) Decode(str string) error {
+	// We must check whether the string was encoded in a URL-safe way in order
+	// to use the appropriate encoding.
+	var err error
+	if strings.ContainsAny(str, "-_") {
```
erikjohnston: *screams*

Why is this required? Don't we specify the base64 encoding? Is this to avoid the need for two different `Base64String` types? Are you just keeping it the way you found it because who knows what will break otherwise?

keyring.go
```diff
@@ -17,6 +17,38 @@ type PublicKeyRequest struct {
...
 }
 
+// PublicKeyNotExpired is a magic value for PublicKeyLookupResult.ExpiredTS:
+// it indicates that this is an active key which has not yet expired
+const PublicKeyNotExpired = Timestamp(0)
+
+// PublicKeyNotValid is a magic value for PublicKeyLookupResult.ValidUntilTS:
+// it is used when we don't have a validity period for this key. Most likely
+// it is an old key with an expiry date.
+const PublicKeyNotValid = Timestamp(0)
```
erikjohnston: Are these two meant to have the same values?

keyring.go
```diff
@@ -123,6 +155,8 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest)
 	k.checkUsingKeys(requests, results, keyIDs, keysFromDatabase)
 
 	for i := range k.KeyFetchers {
+		// TODO: we should distinguish here between expired keys, and those we don't have.
+		// If the key has expired, it's no use re-requesting it.
```
erikjohnston: Reminder to maybe add issue before this merges

keyring.go
```diff
@@ -310,19 +347,48 @@ func (d *DirectKeyFetcher) FetchKeys(
...
-	for req, keys := range results {
+	results := map[PublicKeyRequest]PublicKeyLookupResult{}
+	for req, keys := range serverKeys {
 		// Check that the keys are valid for the server.
 		checks, _, _ := CheckKeys(req.ServerName, time.Unix(0, 0), keys, nil)
 		if !checks.AllChecksOK {
 			return nil, fmt.Errorf("gomatrixserverlib: key response direct from %q failed checks", serverName)
 		}
+
+		mapServerKeysToPublicKeyLookupResult(keys, results)
```
erikjohnston: I'm confused, `mapServerKeysToPublicKeyLookupResult` seems to add multiple keys to the results map, while we didn't before? Does `mapServerKeysToPublicKeyLookupResult` add results under different keys?

base64.go
richvdh: Well, I was deliberately trying to avoid changing too much at once, so yes I was just preserving the original code.

It looks like @NegativeMjark and @babolivier have already been down this path: https://github.com/matrix-org/gomatrixserverlib/pull/58#pullrequestreview-61246401

keyring.go
richvdh: it's fine that they are the same value: they end up in different fields.

keyring.go
richvdh: ack

keyring.go
richvdh: No, I think the keys of the results map are the same, it's just that mapServerKeysToPublicKeyLookupResult turns the `ServerKeys` structures into `PublicKeyLookupResults`.

Compare the implementation of `LookupServerKeys` (https://github.com/matrix-org/gomatrixserverlib/blob/de5aedb583c4a3c26074421b94e4f5c1ac488cf2/client.go#L229) with that of `mapServerKeysToPublicKeyLookupResult`. Note that in both cases the results map key is `PublicKeyRequest{serverName, keyID}` for the relevant server/key.

erikjohnston: Ok, so `mapServerKeysToPublicKeyLookupResult` will insert multiple keys into the map each time, which may clobber existing entries, but that's fine because... we're just undoing the mess that `LookupServerKeys` made? And we're just putting our fingers in our ears when it comes to handling same key appearing with different validity periods?

This is quite a confusing flow to be honest, so a comment would be good

keyring.go
erikjohnston: Fair

base64.go
erikjohnston: Fair, though ewwwwwwwwwww

keyring.go
richvdh: You're right, it was confusing. 8e95683 changes `LookupServerKeys` so that it just returns the list: effectively the map-building it was doing is now pulled up a level.

In terms of the same key appearing multiple times: yes it's a mess, but it's no more of a mess than before, and this PR is really just a starting point for clearing up some of this mess.

PR #73, opened 2017-11-20
Use /key/v2/server to get keys direct

richvdh:
... because /key/v2/query is only if you're talking to a perspectives server.

### DIFF
--- a/keyring.go
+++ b/keyring.go
@@ -333,9 +333,9 @@ func (d *DirectKeyFetcher) FetchKeys(
 	}
 
 	results := map[PublicKeyRequest]PublicKeyLookupResult{}
-	for server, reqs := range byServer {
+	for server := range byServer {
 		// TODO: make these requests in parallel
-		serverResults, err := d.fetchKeysForServer(ctx, server, reqs)
+		serverResults, err := d.fetchKeysForServer(ctx, server)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


keyring.go
```diff
@@ -348,25 +348,24 @@ func (d *DirectKeyFetcher) FetchKeys(
...
-		}
 
-		// TODO: What happens if the same key ID appears in multiple responses?
-		// We should probably take the response with the highest valid_until_ts.
-		mapServerKeysToPublicKeyLookupResult(keys, results)
-	}
+	// TODO (matrix-org/dendrite#345): What happens if the same key ID
+	// appears in multiple responses? We should probably take the response
+	// with the highest valid_until_ts.
+	mapServerKeysToPublicKeyLookupResult(keys, results)
```
erikjohnston: Is this still true if we're only calling this once?

richvdh: yes, the server could still return two keys with the same ID. Though maybe we should reject the whole response in that case?

erikjohnston: It feels like something has gone wrong somewhere along the lines if that happens, but I don't really mind much what we do.

richvdh: To be fair this comment probably applies more to the perspectives case, which I forgot to update; done in 886000e.

PR #74, opened 2017-11-20
Respect result from VerifyEventSignatures

richvdh:
Don't swallow the error when verification fails

### DIFF
--- a/federationtypes.go
+++ b/federationtypes.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"github.com/matrix-org/util"
 )
 
 // A RespSend is the content of a response to PUT /_matrix/federation/v1/send/{txnID}/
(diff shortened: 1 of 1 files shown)


richvdh:
wtf does "file is not goimported" mean?

erikjohnston:
Possibly that the imports are in the wrong order?

federationtypes.go
```diff
@@ -134,8 +136,10 @@ func (r RespState) Check(ctx context.Context, keyRing JSONVerifier) error {
 	}
 
 	// Check if the events pass signature checks.
+	logger.Infof("Checking event signatures for %d events of room state", len(allEvents))
 	if err := VerifyEventSignatures(ctx, allEvents, keyRing); err != nil {
-		return nil
+		logger.Warnf("Event signature check failed: %s", err)
```
erikjohnston: TBH I'd expect the caller to handle/log errors

richvdh: fair. removed.

PR #76, opened 2017-11-20
Add FetcherName() method to KeyFetcher

richvdh:
... and then use it in some logging

### DIFF
--- a/keyring.go
+++ b/keyring.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/matrix-org/util"
 	"golang.org/x/crypto/ed25519"
 )
 
@@ -60,6 +61,10 @@ type KeyFetcher interface {
 	// The result may have more (server name, key ID) pairs than were in the request.
 	// Returns an error if there was a problem fetching the keys.
 	FetchKeys(ctx context.Context, requests map[PublicKeyRequest]Timestamp) (map[PublicKeyRequest]PublicKeyLookupResult, error)
+
+	// FetcherName returns the name of this fetcher, which can then be used for
+	// logging errors etc.
+	FetcherName() string
 }
 
 // A KeyDatabase is a store for caching public keys.
@@ -113,6 +118,7 @@ type JSONVerifier interface {
 
 // VerifyJSONs implements JSONVerifier.
 func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest) ([]VerifyJSONResult, error) { // nolint: gocyclo
+	logger := util.GetLogger(ctx)
 	results := make([]VerifyJSONResult, len(requests))
 	keyIDs := make([][]KeyID, len(requests))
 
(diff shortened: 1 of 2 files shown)


keyring.go
```diff
@@ -163,12 +169,14 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest)
...
 		}
+		logger.Infof("Requesting %d keys from %s", len(keyRequests), fetcher.FetcherName())
 		// TODO: Coalesce in-flight requests for the same keys.
 		// Otherwise we risk spamming the servers we query the keys from.
-		keysFetched, err := k.KeyFetchers[i].FetchKeys(ctx, keyRequests)
+		keysFetched, err := fetcher.FetchKeys(ctx, keyRequests)
 		if err != nil {
 			return nil, err
 		}
+		logger.Infof("Got %d keys from fetcher: verifying", len(keysFromDatabase))
```
erikjohnston: I think maybe we should try and use `WithFields` for things like this, to make it easier to navigate logs (if we start using non-file based logging)

Also in that vein: it might be good to include the `FetcherName` in all log entries? I think you can just do the following to get it included in all log entries:

```golang
logger := util.GetLogger(ctx).WithFields(log.Fields{
    "fetcher_name": fetcher.FetcherName(),
})
```

erikjohnston: OTOH, I guess `FetcherName` isn't the most useful of bits of information to search from, but I think maybe we should just get into the habit? Thoughts?

richvdh: bleh. I think that once I get into fetchers that delegate to other fetchers (as my forthcoming perspectives thing will), it's going to be more confusing than helpful.

erikjohnston: Is that about adding `FletcherName` to this particular log line, or when including it using `WithFields`

I'm tempted to have a policy of never formatting log lines and instead use `WithFields` for any dynamic content. (This also has the nice side effect of being able to easily grep for a specfic log line)

richvdh: oh I see. I thought you meant creating a whole new logger for the context. Yes, I'll have a go at that.

keyring.go
```diff
@@ -163,12 +169,22 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest)
...
 			return results, nil
 		}
+		fetcherLogger := logger.WithField("fetcher", fetcher.FetcherName())
+
 		// TODO: Coalesce in-flight requests for the same keys.
 		// Otherwise we risk spamming the servers we query the keys from.
-		keysFetched, err := k.KeyFetchers[i].FetchKeys(ctx, keyRequests)
+
+		fetcherLogger.WithField("num_key_requests", len(keyRequests)).
+			Info("Requesting keys from fetcher")
```
erikjohnston: I've never seen anyone use a trailing dot

richvdh: as discussed: it doesn't work if you bump the dot to the next line

PR #77, opened 2017-11-21
Make event parsing faster

erikjohnston:


json.go
```diff
@@ -16,66 +16,74 @@
...
 func CanonicalJSON(input []byte) ([]byte, error) {
-	sorted, err := SortJSON(input, make([]byte, 0, len(input)))
-	if err != nil {
-		return nil, err
+	if !gjson.Valid(string(input)) {
+		return nil, errors.Errorf("invalid json")
 	}
-	return CompactJSON(sorted, make([]byte, 0, len(sorted))), nil
+
+	input = CompactJSON(input, make([]byte, 0, len(input)))
```
richvdh: delegate to `CanonicalJSONAssumeValid` here?

event.go
```diff
@@ -183,37 +185,52 @@ func (eb *EventBuilder) Build(eventID string, now time.Time, origin ServerName,
...
 		}
-	}
 
-	if eventJSON, err = CanonicalJSON(eventJSON); err != nil {
-		return
+		redactedJSON = CanonicalJSONAssumeValid(redactedJSON)
+
+		// We need to ensure that `result` is the redacted event.
+		// If redactedJSON is the same as eventJSON then `result` is already
+		// correct. If not then we need to repase.
```
richvdh: repase

event_test.go
```diff
@@ -0,0 +1,52 @@
...
+
+import (
+	"encoding/json"
+	"testing"
+)
+
+func benchmarkParse(b *testing.B, eventJSON string) {
+	var event Event
+
+	// run the Fib function b.N times
```
richvdh: s/Fib/Unmarshal/

json.go
```diff
@@ -16,66 +16,74 @@
...
+	return SortJSON(input, make([]byte, 0, len(input))), nil
+}
+
+// CanonicalJSONAssumeValid is the same as CanonicalJSON, but assumes the
+// input is valid JSON
+func CanonicalJSONAssumeValid(input []byte) []byte {
+	input = CompactJSON(input, make([]byte, 0, len(input)))
```
richvdh: I'm wondering why we used to Sort and then Compact, and now we Compact and then Sort. It looks like SortJSON removes extraneous whitespace, so perhaps it's more efficient to Compact after Sorting? (In fact, should we combine the two by doing string compaction when the input is a string in `sortJSONValue`?)

erikjohnston: To be honest, it was mainly that it made it easier debugging for various spurious reasons. I don't think it matters one way or the other.

> In fact, should we combine the two by doing string compaction when the input is a string in sortJSONValue

Quite possibly, canonicalisation takes about 10%

PR #81, opened 2017-11-30
Add Event.SetUnsignedField

erikjohnston:


event.go
```diff
@@ -306,6 +307,33 @@ func (e Event) SetUnsigned(unsigned interface{}) (Event, error) {
...
+// path is a dot separated path into the unsigned dict (see gjson package
+// for details on format). In particular some characters like '.' and '*' must
+// be escaped.
+func (e *Event) SetUnsignedField(path string, value interface{}) error {
+	// The safest way is to change the unsigned json and then reparse the
+	// event fully. But since we are only changing the unsigned section,
+	// which doesn't affect the signatures or hashes, we can cheat and
+	// just fiddle those bits directly.
+
+	path = fmt.Sprintf("unsigned.%s", path)
```
richvdh: `"unsigned." + path`, surely?

event.go
```diff
@@ -306,6 +307,33 @@ func (e Event) SetUnsigned(unsigned interface{}) (Event, error) {
...
+	eventJSON, err := sjson.SetBytes(e.eventJSON, path, value)
+	if err != nil {
+		return err
+	}
+	eventJSON = CanonicalJSONAssumeValid(eventJSON)
```
richvdh: can we avoid re-canonicalising the whole thing by canonicalising the unsigned bit before doing `sjson.SetBytes`?

erikjohnston: If the `unsigned` key didn't exist then it might get added out of order. I think its generally just safer to re-canonicalise the entire thing tbh

erikjohnston: (and I don't think canoicalising is expensive enough here to worry, though I could be wrong.)

event_test.go
```diff
@@ -50,3 +50,32 @@ func BenchmarkParseSmallerEventFailedHash(b *testing.B) {
...
+	if err := json.Unmarshal([]byte(initialEventJSON), &event); err != nil {
+		t.Error("Failed to parse event")
+	}
+
+	err := event.SetUnsignedField("foo", "bar")
+	if err != nil {
+		t.Error("Failed to insert foo")
+	}
+
+	event.SetUnsignedField("x", 1)
```
richvdh: err :=

PR #83, opened 2017-12-08
Fix timeouts for federationclient

richvdh:
Construct the Client used by FederationClient correctly, so that it is
configured to time out requests.


erikjohnston:
Modulo it not building

PR #85, opened 2017-12-15
Improve logging of outgoing federation requests

richvdh:
Try to log outbound federation requests more consistently and more usefully

### DIFF
--- a/client.go
+++ b/client.go
@@ -332,18 +332,24 @@ func (fc *Client) DoRequestAndParseResponse(
 //
 func (fc *Client) DoHTTPRequest(ctx context.Context, req *http.Request) (*http.Response, error) {
 	reqID := util.RandomString(12)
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -332,18 +332,24 @@ func (fc *Client) DoRequestAndParseResponse(
 //
 func (fc *Client) DoHTTPRequest(ctx context.Context, req *http.Request) (*http.Response, error) {
 	reqID := util.RandomString(12)
-	logger := util.GetLogger(ctx).WithField("server", req.URL.Host).WithField("out.req.ID", reqID)
+	logger := util.GetLogger(ctx).
+		WithField("out.req.ID", reqID).
```
erikjohnston: Normally one would use

```golang
logger.WithFields(logrus.Fields{
   "out.req.ID": reqID.
   ...
})
```

richvdh: so one would

PR #92, opened 2018-06-02
Add AS dummy device ID

APwhitehat:
Required to have a common device ID for all AS.

`Signed-off-by: Anant Prakash <anantprakashjsr@gmail.com>`


anoadragon453:
Might want to stick in [appservice.go](https://github.com/matrix-org/gomatrixserverlib/blob/master/appservice.go) btw.

APwhitehat:
@anoadragon453 yeah. Should have fetched upstream before making a new file.

anoadragon453:
Looking at this again, I'm not sure if this is something that belongs in gomatrixserverlib, only because I'm not sure other matrix go server implementation really need this?

APwhitehat:
closed in favour of  matrix-org/dendrite#487

PR #95, opened 2018-06-28
Add missing prev_state field

anoadragon453:
WIP for https://github.com/matrix-org/dendrite/issues/519

Allows `30room-join.pl` to pass.

This field has been deprecated, and Synapse doesn't even use it, but it does check that it exists.

Put in a TODO for us to hopefully remove it in the future on a spec upgrade.

### DIFF
--- a/event.go
+++ b/event.go
@@ -61,6 +61,8 @@ type EventBuilder struct {
 	StateKey *string `json:"state_key,omitempty"`
 	// The events that immediately preceded this event in the room history.
 	PrevEvents []EventReference `json:"prev_events"`
+	// TODO: A deprecated, empty but yet still expected key.
+	PrevState string `json:"prev_state"`
 	// The events needed to authenticate this event.
 	AuthEvents []EventReference `json:"auth_events"`
 	// The event ID of the event being redacted if this event is a "m.room.redaction".


event.go
```diff
@@ -61,6 +61,8 @@ type EventBuilder struct {
 	StateKey *string `json:"state_key,omitempty"`
 	// The events that immediately preceded this event in the room history.
 	PrevEvents []EventReference `json:"prev_events"`
+	// TODO: A deprecated, empty but yet still expected key.
+	PrevState string `json:"prev_state"`
```
erikjohnston: This should be a list, and should be omitted if not a state event

anoadragon453:
Any idea how we "omitempty" when we should have that key normally, but 
it should be an empty array?

anoadragon453:
So an empty array isn't affected by `omitempty`: https://play.golang.org/p/yPx3b6L09CH

But I think I'm going to leave this one alone for now, merely because the way this method is set up is utterly confusing to me :/

APwhitehat:
@anoadragon453 https://play.golang.org/p/eUdWgTFjhTl
json:omitempty only omits if it detects a default value or a nil

anoadragon453:
Obsoleted by https://github.com/matrix-org/sytest/pull/642

PR #96, opened 2018-07-04
Remove AS-specific event type in favor of generic Event

anoadragon453:
We were using an event type that was specific for AS's, but this is kind of a pain because whenever we add something important to the event type, you'd also have to update this event type so that AS's got the new info as well.

Instead, just use the same type for everything.

### DIFF
--- a/appservice.go
+++ b/appservice.go
@@ -15,27 +15,8 @@
 
 package gomatrixserverlib
 
-// ApplicationServiceUnsigned is the contents of the unsigned field of an
-// ApplicationServiceEvent.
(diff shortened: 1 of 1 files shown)


t3chguy:
This was discussed in `#dendrite-dev:matrix.org`

anoadragon453:
Specifically the debate was about `user_id` no longer being included in events sent to AS's, however this is already planned for r0, and the AS docs are just outdated (we're getting to that, though!).

turt2live:
ref: https://github.com/matrix-org/matrix-doc/issues/1269

PR #101, opened 2018-11-08
Add support for backfill requests

babolivier:


federationclient.go
```diff
@@ -170,3 +171,30 @@ func (ac *FederationClient) LookupRoomAlias(
...
+	limitStr := strconv.Itoa(limit)
+
+	// Define the URL's query.
+	query := url.Values{}
+	query["v"] = eventIDs
+	query.Set("limit", limitStr)
+
+	// Use the url.URL structure to easily generate the request's URI (path?query).
+	u := url.URL{
+		Path:     "/_matrix/federation/v1/backfill/" + roomID,
```
erikjohnston: Does this not need a trailing slash? And to url encode the roomID?

babolivier: Does need a trailing slash indeed. roomID is automatically url encoded when calling `u.RequestURI()`, as well as the event IDs.

babolivier: Trailing slash fixed in https://github.com/matrix-org/gomatrixserverlib/pull/101/commits/6802ea9d837d411125f79c3a1e254022074726f3.

babolivier: Hmm. I didn't think of that. Fixed in 9682b8b.

erikjohnston: Fair, its one of the classic mistakes literally *everyone* makes :)

PR #104, opened 2019-01-30
Add functions for performing a well-known lookup on a matrix server

anoadragon453:


well_known.go
```diff
@@ -0,0 +1,51 @@
...
+	NewAddress ServerName `json:"m.server"`
+}
+
+// LookupWellKnown looks up a well-known record for a matrix server. If one if
+// found, it returns the server to redirect to.
+func LookupWellKnown(serverNameType ServerName) (*WellKnownResult, error) {
+	serverName := string(serverNameType)
+
+	// Handle ending "/"
+	strings.Trim(serverName, "/")
```
richvdh: surely it is an error for a server name to include a / ? we should reject it, not trim it.

PR #105, opened 2019-02-01
Shorten line length of method declaration

anoadragon453:
~~Adds parameters to `CheckKeys` that lets you specify whether well-known is being used or not, and the new server name if so.~~

~~https://github.com/matrix-org/matrix-federation-tester/pull/19~~

Simply does some aesthetics updates now as all PR functionality was moved to matrix-federation-tester.

### DIFF
--- a/keys.go
+++ b/keys.go
@@ -161,7 +161,7 @@ type KeyChecks struct {
 }
 
 // CheckKeys checks the keys returned from a server to make sure they are valid.
-// If the checks pass then also return a map of key_id to Ed25519 public key and a list of SHA256 TLS fingerprints.
+// If the checks pass then also return a map of key_id to Ed25519 public key
 func CheckKeys(
 	serverName ServerName,
 	now time.Time,
--- a/keys.go
+++ b/keys.go
@@ -166,12 +166,11 @@ func CheckKeys(
 	serverName ServerName,
 	now time.Time,
 	keys ServerKeys,
-	wellKnown *WellKnownResult,
 ) (
 	checks KeyChecks,
 	ed25519Keys map[KeyID]Base64String,
 ) {
-	checks.MatchingServerName = serverName == keys.ServerName || serverName == wellKnown.NewAddress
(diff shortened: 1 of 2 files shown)


keys.go
```diff
@@ -173,10 +173,18 @@ type KeyChecks struct {
...
+	keys ServerKeys,
+	connState *tls.ConnectionState,
+	wellKnown *WellKnownResult,
+) (
+	checks KeyChecks,
+	ed25519Keys map[KeyID]Base64String,
+	sha256Fingerprints []Base64String,
 ) {
-	checks.MatchingServerName = serverName == keys.ServerName
+	checks.MatchingServerName = serverName == keys.ServerName || serverName == wellKnown.NewAddress
```
richvdh: in any case: the returned server name should *never* be the NewAddress returned by the well-known.

anoadragon453:
Un-done weirdness with this check, thanks for the explanation.

I've now broken up serverName and serverHost into two separate vars, though currently none of the checks really care (afaict) if the serverName is different from the matrix server's host. So I've left the breakup there in case we need a distinction in the future. (If you think this is weird I can just remove it).

keys.go
```diff
@@ -161,9 +161,14 @@ type KeyChecks struct {
 }
 
 // CheckKeys checks the keys returned from a server to make sure they are valid.
-// If the checks pass then also return a map of key_id to Ed25519 public key
-func CheckKeys(serverName ServerName, now time.Time, keys ServerKeys) (
-	checks KeyChecks, ed25519Keys map[KeyID]Base64String,
+// If the checks pass then also return a map of key_id to Ed25519 public key and a list of SHA256 TLS fingerprints.
```
richvdh: It doesn't return the TLS fingerprints afaict?

anoadragon453:
Erm, yep, that's all that's left of this PR. LGTY?

richvdh:
well, I guess. It seems a bit pointless tbh, but if you want to keep it, change the PR title.

PR #109, opened 2019-02-06
Fix ValidCertificates check not working when using well-known

anoadragon453:
Fixes https://github.com/matrix-org/matrix-federation-tester/issues/26
Fixes https://github.com/matrix-org/matrix-federation-tester/issues/27

Was due to port (coming from well-known JSON) being included in the DNSName field when checking certificate validity. This of course doesn't make sense (ports are not part of domain names).

Chop off everything after a `:` if it exists. Should be fine as we're not processing any IP(v6)s here.


certificates.go
```diff
@@ -2,13 +2,18 @@ package gomatrixserverlib
...
 import (
 	"crypto/x509"
+	"strings"
 )
 
 // IsValidCertificate checks if the given x509 certificate can be verified using
 // system root CAs and an optional pool of intermediate CAs.
 func IsValidCertificate(serverName ServerName, c *x509.Certificate, intermediates *x509.CertPool) (valid bool, err error) {
+	// Remove port from serverName if it exists
+	serverNameCleaned := strings.Split(string(serverName), ":")[0]
```
richvdh: this won't work if the servername is an ipv6 literal.

(also will it do the correct thing if there is no port?)

anoadragon453: Surely this check should just fail if you have an ipv6 literal, yes?

anoadragon453: Will do correct thing if no port: https://play.golang.org/p/rsUl7ElmqY2

richvdh: errrm why should it fail if we're using an ip literal?

anoadragon453: Are we supposed TLS certificates for IP literals? (is this a thing?)

anoadragon453: https://1.1.1.1/ oh god it is

anoadragon453: Ok, this seems to do the trick: https://play.golang.org/p/KzGn5PmiKaI

richvdh:
In fact you've already written the tests for the playground. Go has a shiny test framework. go use it ;-p

richvdh:
> "1fff:0:a88:85a3::ac1f"

^^ not a valid server name

anoadragon453:
> > "1fff:0:a88:85a3::ac1f"
> 
> ^^ not a valid server name

Oh, they can't be in short form?

anoadragon453:
Nevermind, https://matrix.org/docs/spec/appendices.html#server-name explains it.

anoadragon453:
So one thing is I'm not entirely sure whether an IP address can go into `x509.VerifyOptions.DNSName`. My line of thinking just now is to error on an IP address and tell the user we don't currently support certificate validation checked with IP address serverNames.

What do you think?

certificates.go
```diff
@@ -2,16 +2,45 @@ package gomatrixserverlib
...
 	"crypto/x509"
+	"errors"
+	"net"
+	"strings"
 )
 
 // IsValidCertificate checks if the given x509 certificate can be verified using
 // system root CAs and an optional pool of intermediate CAs.
 func IsValidCertificate(serverName ServerName, c *x509.Certificate, intermediates *x509.CertPool) (valid bool, err error) {
+	// Clean and verify serverName
```
richvdh: pointless comment is pointless

certificates.go
```diff
@@ -2,16 +2,45 @@ package gomatrixserverlib
...
 	}
 	roots, err := c.Verify(verificationOpts)
 
 	return len(roots) > 0, err
 }
+
+// cleanServerName is a function that takes in a ServerName, verifies it, and
+// removes the port if necessary
+// TODO: Support IP Addresses (and update test)
+func cleanAndVerifyServerName(serverName ServerName) (serverNameCleaned string, err error) {
```
richvdh: it doesn't seem to do any verifying?

certificates_test.go
```diff
@@ -0,0 +1,51 @@
+/* Copyright 2017 Vector Creations Ltd
```
richvdh: it's 2019, and we're not called Vector Creations any more

richvdh:
@anoadragon453 : sorry, couldn't resist playing with some go. I've given this a bit of a substantial refactor, as you may be able to see.

I also had a look at the source for the certificate verification (https://golang.org/src/crypto/x509/verify.go#L1000) and it looks like it should handle IP addresses just fine, so I've removed that check.

LMK what you think

(have just seen the CI is failing: have to step away for now but will try to sort it out shortly)

richvdh:
incidentally, as pointed out in https://github.com/matrix-org/matrix-federation-tester/pull/35, there is a [net.SplitHostPort](https://golang.org/pkg/net/#SplitHostPort), which might have been useful here; however, it doesn't handle the case that a port is optional, so I'm not sure we'd actually end up saving anything by attempting to use it.

certificates.go
```diff
@@ -2,13 +2,21 @@ package gomatrixserverlib
...
 	"crypto/x509"
+	"errors"
 )
 
 // IsValidCertificate checks if the given x509 certificate can be verified using
 // system root CAs and an optional pool of intermediate CAs.
 func IsValidCertificate(serverName ServerName, c *x509.Certificate, intermediates *x509.CertPool) (valid bool, err error) {
+	host, _, isValid := ParseAndValidateServerName(serverName)
+	if !isValid {
+		return false, errors.New("Not a valid serverName")
```
anoadragon453: I'm planning to have these errors appear in the output JSON and thus be human-readable. Can we print out which serverName it's validating in this error for easier debugging?

richvdh: done

anoadragon453:
LGTM if it does it you.

*waits for CI*

federationtypes.go
```diff
@@ -4,10 +4,100 @@ import (
...
+	if host[0] == '[' {
+		// must be a valid IPv6 address
+		if host[len(host)-1] != ']' {
+			return
+		}
+		ip := host[1 : len(host)-1]
+		if net.ParseIP(ip) == nil {
+			return
+		}
+	} else if unicode.IsDigit(rune(host[0])) {
```
richvdh: gah; this is completely bogus. Hostnames can begin with digits.

anoadragon453: Oh poo, they can: https://0net.io/

PR #111, opened 2019-02-08
Allow valid certificate checker to check with multiple DNSNames

anoadragon453:
Paired with: https://github.com/matrix-org/matrix-federation-tester/pull/33


anoadragon453:
Doesn't seem like this is necessary anymore.

PR #112, opened 2019-02-13
Include a descriptive error in WellKnownLookup

anoadragon453:
For https://github.com/matrix-org/matrix-federation-tester/pull/42

Include an error as part of the WellKnownLookup struct which can be presented in matrix-federation-tester.

WellKnownLookup changed from a pointer to a non-pointer to match stuff in matrix-federation-tester. Not sure if bad.

### DIFF
--- a/well_known.go
+++ b/well_known.go
@@ -12,7 +12,7 @@ import (
 // Located at https://<server_name>/.well-known/matrix/server
 type WellKnownResult struct {
 	NewAddress ServerName `json:"m.server"`
-	Error string `json:"Error,omitempty"`
+	Error      string     `json:"Error,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


well_known.go
```diff
@@ -12,11 +12,12 @@ import (
 // Located at https://<server_name>/.well-known/matrix/server
 type WellKnownResult struct {
 	NewAddress ServerName `json:"m.server"`
+	Error      string     `json:"Error,omitempty"`
```
richvdh: this doesn't feel like the right thing to be doing. `WellKnownResult` is supposed to represent the thing that gets returned from the server. Surely we shouldn't be adding extra fields to it? Make another structure if need be.

anoadragon453:
Turns out doing it the other way requires no changes, yay.

PR #113, opened 2019-02-13
Fix hostname validation bug

richvdh:
Fixes a bug where hostnames starting with digits were rejected


anoadragon453:
Apparently a domain name can't start or end with `-`, but do we care?

richvdh:
hrm, not sure I care particularly.

PR #115, opened 2019-02-15
Make FetchKeysDirect generate nicer human-readable messages so they display on the federation tester in a nicer way

babolivier:
Fixes https://github.com/matrix-org/matrix-federation-tester/issues/41 and https://github.com/matrix-org/matrix-federation-tester/issues/40

### DIFF
--- a/keys.go
+++ b/keys.go
@@ -19,6 +19,7 @@ import (
 	"bufio"
 	"crypto/tls"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"net"
 	"net/http"
--- a/keys.go
+++ b/keys.go
@@ -134,7 +134,7 @@ func FetchKeysDirect(serverName ServerName, addr, sni string) (*ServerKeys, *tls
 		return nil, nil, err
 	}
 	if response.StatusCode != http.StatusOK {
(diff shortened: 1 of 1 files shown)


keys.go
```diff
@@ -132,9 +133,12 @@ func FetchKeysDirect(serverName ServerName, addr, sni string) (*ServerKeys, *tls
 	if err != nil {
 		return nil, nil, err
 	}
+	if response.StatusCode != http.StatusOK {
+		return nil, nil, fmt.Errorf("Remote host responded with non-OK status when attempting to fetch server keys: %s", response.Status)
```
richvdh: I think this could be clearer/more succinct:

```suggestion
		return nil, nil, fmt.Errorf("Non-200 response %d from remote server", response.StatusCode)
```

keys.go
```diff
@@ -132,9 +133,12 @@ func FetchKeysDirect(serverName ServerName, addr, sni string) (*ServerKeys, *tls
...
+	}
 	var keys ServerKeys
 	if err = json.NewDecoder(response.Body).Decode(&keys); err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("Couldn't decode JSON when attempting to fetch server keys: %s", err.Error())
```
richvdh:
```suggestion
		return nil, nil, errors.Wrap(err, "Unable to decode JSON from remote server")
```

PR #116, opened 2019-02-18
Don't error on lack of .well-known

babolivier:
Add a boolean value which describes whether a `/.well-known/matrix/server` file could be found for this server, instead of displaying the lack of one as an error.

Required by https://github.com/matrix-org/matrix-federation-tester/pull/52

### DIFF
--- a/well_known.go
+++ b/well_known.go
@@ -2,7 +2,6 @@ package gomatrixserverlib
 
 import (
 	"encoding/json"
-	"errors"
 	"io/ioutil"
 	"net/http"
 	"strings"
@@ -16,7 +15,7 @@ type WellKnownResult struct {
 
 // LookupWellKnown looks up a well-known record for a matrix server. If one if
 // found, it returns the server to redirect to.
-func LookupWellKnown(serverNameType ServerName) (*WellKnownResult, error) {
+func LookupWellKnown(serverNameType ServerName) (*WellKnownResult, bool, error) {
 	serverName := string(serverNameType)
 
 	// Handle ending "/"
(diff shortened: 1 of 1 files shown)


well_known.go
```diff
@@ -16,7 +15,7 @@ type WellKnownResult struct {
 
 // LookupWellKnown looks up a well-known record for a matrix server. If one if
 // found, it returns the server to redirect to.
-func LookupWellKnown(serverNameType ServerName) (*WellKnownResult, error) {
+func LookupWellKnown(serverNameType ServerName) (*WellKnownResult, bool, error) {
```
Half-Shot: Passing though: Why a new boolean instead of `null, null`. A null error and a null WellKnownResult would imply that no well known was found, but it's also not an error?

babolivier: fwiw I did it that way because imho it makes the code easier to understand, but maybe the same thing could be achieved through returning `nil, nil` and documenting this behaviour in the doc comment on top of the function? Not sure which one is best here.

Half-Shot: It's personal taste, and I understand the reason. I think I'd put the boolean before the WellKnownResult but that might not be how Go prefers it. Feel free to ignore me :)

well_known.go
```diff
@@ -28,27 +28,27 @@ func LookupWellKnown(serverNameType ServerName) (*WellKnownResult, error) {
 	// Request server's well-known record
 	resp, err := http.Get(wellKnown)
 	if err != nil {
-		return nil, err
+		return nil, false, err
```
richvdh: why does this return an err, but a non-200 response does not? they both amount to the same thing: that the well-known request was unsuccessful.

The problem is not that LookupWellKnown returns an error when there is no well-known. The problem is that the federation tester takes the error and sticks it straight into a field called 'error' in the json.

PR #118, opened 2019-02-21
Add support for .well-known when resolving a server name

### LINKED ISSUE #117: Outbound federation requests should take .well-known into account
We already have a function for calculating a server's .well-known. This should be used whenever making a new request to a server, cached, and then used for all subsequent requests.

babolivier:
Resolves #117

### DIFF
--- a/client.go
+++ b/client.go
@@ -101,19 +101,19 @@ func makeHTTPSURL(u *url.URL, addr string) (httpsURL url.URL) {
 
 func (f *federationTripper) RoundTrip(r *http.Request) (*http.Response, error) {
 	serverName := ServerName(r.URL.Host)
-	dnsResult, err := LookupServer(serverName)
+	hosts, err := ResolveServer(serverName, true)
 	if err != nil {
 		return nil, err
 	}
 
-	if len(dnsResult.Addrs) == 0 {
+	if len(hosts) == 0 {
 		return nil, fmt.Errorf("no address found for matrix host %v", serverName)
 	}
 
 	var resp *http.Response
 	// TODO: respect the priority and weight fields from the SRV record
-	for _, addr := range dnsResult.Addrs {
-		u := makeHTTPSURL(r.URL, addr)
+	for _, host := range hosts {
+		u := makeHTTPSURL(r.URL, host)
 		r.URL = &u
 		resp, err = f.transport.RoundTrip(r)
 		if err == nil {
--- a/client.go
+++ b/client.go
@@ -22,7 +22,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io/ioutil"
-	"net"
 	"net/http"
 	"net/url"
 	"strings"
@@ -67,26 +66,9 @@ func newFederationTripper() *federationTripper {
 	// TODO: Verify ceritificates
 	return &federationTripper{
 		transport: &http.Transport{
-			// Set our own DialTLS function to avoid the default net/http SNI.
-			// By default net/http and crypto/tls set the SNI to the target host.
-			// By avoiding the default implementation we can keep the ServerName
-			// as the empty string so that crypto/tls doesn't add SNI.
-			DialTLS: func(network, addr string) (net.Conn, error) {
-				rawconn, err := net.Dial(network, addr)
-				if err != nil {
-					return nil, err
-				}
-				// Wrap a raw connection ourselves since tls.Dial defaults the SNI
-				conn := tls.Client(rawconn, &tls.Config{
-					ServerName: "",
-					// TODO: We should be checking that the TLS certificate we see here matches
-					//       one of the allowed SHA-256 fingerprints for the server.
-					InsecureSkipVerify: true, // nolint: gas
-				})
-				if err := conn.Handshake(); err != nil {
-					return nil, err
-				}
-				return conn, nil
+			TLSClientConfig: &tls.Config{
+				// TODO: Remove this when we enforce MSC1711.
+				InsecureSkipVerify: true,
 			},
 		},
 	}
@@ -101,20 +83,22 @@ func makeHTTPSURL(u *url.URL, addr string) (httpsURL url.URL) {
 
 func (f *federationTripper) RoundTrip(r *http.Request) (*http.Response, error) {
 	serverName := ServerName(r.URL.Host)
-	hosts, err := ResolveServer(serverName, true)
+	resolutionResults, err := ResolveServer(serverName)
 	if err != nil {
 		return nil, err
 	}
 
-	if len(hosts) == 0 {
+	if len(resolutionResults) == 0 {
 		return nil, fmt.Errorf("no address found for matrix host %v", serverName)
 	}
 
 	var resp *http.Response
 	// TODO: respect the priority and weight fields from the SRV record
-	for _, host := range hosts {
-		u := makeHTTPSURL(r.URL, host)
+	for _, result := range resolutionResults {
+		u := makeHTTPSURL(r.URL, result.Destination)
 		r.URL = &u
+		r.Host = string(result.Host)
+		f.transport.(*http.Transport).TLSClientConfig.ServerName = result.Name
 		resp, err = f.transport.RoundTrip(r)
 		if err == nil {
 			return resp, nil
--- a/client.go
+++ b/client.go
@@ -59,19 +59,30 @@ func NewClientWithTimeout(timeout time.Duration) *Client {
 }
 
 type federationTripper struct {
-	transport http.RoundTripper
+	transports map[string]http.RoundTripper
 }
 
 func newFederationTripper() *federationTripper {
-	// TODO: Verify ceritificates
-	return &federationTripper{
-		transport: &http.Transport{
+	tripper := new(federationTripper)
+	tripper.transports = make(map[string]http.RoundTripper)
+	return tripper
+}
+
+// getTransport returns a http.Transport instance with a TLS configuration using
+// the given server name for SNI. It also creates the instance if there isn't
+// any for this server name.
+func (f *federationTripper) getTransport(tlsServerName string) http.RoundTripper {
+	if transport, ok := f.transports[tlsServerName]; !ok || transport == nil {
+		f.transports[tlsServerName] = &http.Transport{
 			TLSClientConfig: &tls.Config{
+				ServerName: tlsServerName,
 				// TODO: Remove this when we enforce MSC1711.
 				InsecureSkipVerify: true,
 			},
-		},
+		}
 	}
+
+	return f.transports[tlsServerName]
 }
 
 func makeHTTPSURL(u *url.URL, addr string) (httpsURL url.URL) {
@@ -98,8 +109,7 @@ func (f *federationTripper) RoundTrip(r *http.Request) (*http.Response, error) {
 		u := makeHTTPSURL(r.URL, result.Destination)
 		r.URL = &u
 		r.Host = string(result.Host)
-		f.transport.(*http.Transport).TLSClientConfig.ServerName = result.Name
-		resp, err = f.transport.RoundTrip(r)
+		resp, err = f.getTransport(result.TLSServerName).RoundTrip(r)
 		if err == nil {
 			return resp, nil
 		}
--- a/client.go
+++ b/client.go
@@ -67,8 +67,7 @@ type federationTripper struct {
 
 func newFederationTripper() *federationTripper {
 	return &federationTripper{
-		transports:      make(map[string]http.RoundTripper),
-		transportsMutex: sync.Mutex{},
+		transports: make(map[string]http.RoundTripper),
 	}
 }
 
--- a/client.go
+++ b/client.go
@@ -62,13 +62,13 @@ func NewClientWithTimeout(timeout time.Duration) *Client {
 type federationTripper struct {
 	// transports maps an TLS server name with an HTTP transport.
 	transports      map[string]http.RoundTripper
-	transportsMutex *sync.Mutex
+	transportsMutex sync.Mutex
 }
 
 func newFederationTripper() *federationTripper {
 	return &federationTripper{
 		transports:      make(map[string]http.RoundTripper),
-		transportsMutex: new(sync.Mutex),
+		transportsMutex: sync.Mutex{},
 	}
 }
 
@@ -76,8 +76,8 @@ func newFederationTripper() *federationTripper {
 // the given server name for SNI. It also creates the instance if there isn't
 // any for this server name.
 // We need to use one transport per TLS server name (instead of giving our round
-// tripper a single transport) because a transport is bound to a single server
-// name.
+// tripper a single transport) because there is no way to specify the TLS
+// ServerName on a per-connection basis.
 func (f *federationTripper) getTransport(tlsServerName string) (transport http.RoundTripper) {
 	var ok bool
 
--- a/client.go
+++ b/client.go
@@ -25,6 +25,7 @@ import (
 	"net/http"
 	"net/url"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/matrix-org/gomatrix"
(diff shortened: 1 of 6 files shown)


resolve.go
```diff
@@ -107,3 +108,67 @@ func LookupServer(serverName ServerName) (*DNSResult, error) { // nolint: gocycl
 
 	return &result, nil
 }
+
+// ResolveServer implements the server name resolution algorithm described at
+// https://matrix.org/docs/spec/server_server/r0.1.1.html#resolving-server-names
+// Returns a slice containing the hosts (using the host:port form) that can be
+// used to send a federation request to the server using a given server name.
+// needWellKnown indicates whether a .well-known file should be looked up.
```
richvdh: why would this be false?

resolve.go
```diff
@@ -107,3 +108,67 @@ func LookupServer(serverName ServerName) (*DNSResult, error) { // nolint: gocycl
...
+// Returns an error if the server name isn't valid, or if either the .well-known
+// lookup or any DNS lookup failed. Doesn't return an error if no .well-known
+// file could be found for the given server name.
+// nolint: gocyclo
```
richvdh: please don't do this

resolve.go
```diff
@@ -107,3 +108,67 @@ func LookupServer(serverName ServerName) (*DNSResult, error) { // nolint: gocycl
 
 	return &result, nil
 }
+
+// ResolveServer implements the server name resolution algorithm described at
+// https://matrix.org/docs/spec/server_server/r0.1.1.html#resolving-server-names
+// Returns a slice containing the hosts (using the host:port form) that can be
+// used to send a federation request to the server using a given server name.
+// needWellKnown indicates whether a .well-known file should be looked up.
+// Returns an error if the server name isn't valid, or if either the .well-known
```
richvdh: failure to parse the wellKnown should not be treated as an error

well_known.go
```diff
@@ -8,6 +8,10 @@ import (
 	"strings"
 )
 
+var (
+	errNoWellKnown = errors.New("No .well-known found")
```
anoadragon453: Could you change this to a `fmt.Errorf`?

babolivier:
> this doesn't seem to set the host/sni correctly according to the `.well-known` ?

meh, I knew I was forgetting something.

resolve.go
babolivier: Oh that's true

well_known.go
babolivier: Not sure I see the benefits of doing that here, but sure I guess?

resolve.go
babolivier: Same reason Synapse does it that way (iirc): when it gets a result for a .well-known lookup, it calls itself with it because the rest of the steps are the same as when we want to resolve a server name, except we must skip the .well-known lookup.

richvdh: oh! hah. yes, sorry, I hadn't seen that.

Still, it would be nice to hide it from the public interface. Maybe make ResolveServer a thin wrapper around `resolveServer` which just sets needWellKnown.

babolivier: Sure, good idea.

babolivier:
This PR adds two third-party libraries to the repo, which are only used in unit tests:

* https://github.com/h2non/gock, which does HTTP request mocking so the `.well-known` tests don't actually rely on the network
* https://github.com/miekg/dns, which is a popular https://github.com/miekg/dns

babolivier:
This PR adds two third-party libraries to the repo, which are only used in unit tests:

* https://github.com/h2non/gock, which does HTTP request mocking so the `.well-known` tests don't actually rely on the network
* https://github.com/miekg/dns, which is a popular DNS libraries I'm using to make DNS tests not rely on the network

client.go
```diff
@@ -101,20 +83,22 @@ func makeHTTPSURL(u *url.URL, addr string) (httpsURL url.URL) {
...
 
 	var resp *http.Response
 	// TODO: respect the priority and weight fields from the SRV record
-	for _, addr := range dnsResult.Addrs {
-		u := makeHTTPSURL(r.URL, addr)
+	for _, result := range resolutionResults {
+		u := makeHTTPSURL(r.URL, result.Destination)
 		r.URL = &u
+		r.Host = string(result.Host)
+		f.transport.(*http.Transport).TLSClientConfig.ServerName = result.Name
```
richvdh: Is it not the case that a given federationTransport can be used for multiple concurrent outbound requests, so this could end up interfering with other requests which happen at the same time?

resolve_test.go
```diff
@@ -0,0 +1,283 @@
...
+// Tests step 3d of the resolution algorithm.
+func TestHostnameWellKnownWithHostnameNoSRV(t *testing.T) {
+	defer gock.Off()
+
+	gock.New("https://" + hostname).
+		Get("/.well-known/matrix/server").
+		Reply(200).
+		JSON(WellKnownResult{NewAddress: delegatedAddress})
+
+	defer clearFakeDNS(setupFakeDNS(false))
```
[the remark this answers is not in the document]
babolivier: > won't this defer the call to `setupFakeDNS` ?

No, the parameters of the deferred function are evaluated when deferring (as opposed to when running the deferred function). I like your solution better, though.

resolve.go
```diff
@@ -107,3 +116,120 @@ func LookupServer(serverName ServerName) (*DNSResult, error) { // nolint: gocycl
...
+
+		return
+	}
+
+	if needWellKnown {
+		// 3. If the hostname is not an IP literal
+		var result *WellKnownResult
+		result, err = LookupWellKnown(serverName)
+		if err == nil {
+			if len(result.NewAddress) > 0 {
```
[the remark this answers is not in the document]
babolivier: Hmm, I thought I read something about that in the spec, but can't find it anymore, so I'll remove the check.

resolve.go
```diff
@@ -107,3 +116,120 @@ func LookupServer(serverName ServerName) (*DNSResult, error) { // nolint: gocycl
...
+	dnsResults, err := LookupServer(serverName)
+	if err != nil {
+		return
+	}
+
+	// We can't check the length of dnsResults.SRVRecords because LookupServer
+	// might have added a serverName:8448 there. Instead we check whether
+	// there's a dnsResult.SRVError.
+	if dnsResults.SRVError == nil {
+		// 4. If the /.well-known request resulted in an error response
```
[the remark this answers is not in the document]
babolivier: Nothing, my bad.

resolve.go
```diff
@@ -107,3 +116,120 @@ func LookupServer(serverName ServerName) (*DNSResult, error) { // nolint: gocycl
...
+		}
+	}
+
+	// handleNoWellKnown implements steps 4 and 5 of the algorithm (as well as
+	// 3.3 and 3.4)
+	return handleNoWellKnown(serverName, host)
+}
+
+func handleNoWellKnown(serverName ServerName, host string) (results []ResolutionResult, err error) {
+	dnsResults, err := LookupServer(serverName)
```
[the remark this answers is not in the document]
babolivier: I agree with that.

resolve_test.go
```diff
@@ -0,0 +1,283 @@
...
+	"fmt"
+	"net"
+	"reflect"
+	"testing"
+
+	"github.com/miekg/dns"
+	"gopkg.in/h2non/gock.v1"
+)
+
+const (
```
richvdh: it might be good to have some tests for ipv6 things here too.

babolivier: Why? Apart from the SRV lookup there's no DNS lookup involved on gomatrixserverlib's side because that being taken care of by the core HTTP library.

client.go
babolivier: Yes, but as far as I'm aware that's currently the only way to set the server name.

resolve_test.go
richvdh: well, sure, but is a server name of `[1::1:1]` correctly handled as an IPv6 literal with no port? Are all three result fields set correctly? 

How about `[ab::cd]:12` ?

client.go
richvdh: I see what you mean :/

I think it's a bit of a show-stopper though, and that we need to find a solution to it. It might be that we need a separate `http.Transport` for each server name?

resolve_test.go
babolivier: Aaah I didn't understand you correctly. Sure, I'll add some.

client.go
babolivier: That sound like a decent solution.

resolve.go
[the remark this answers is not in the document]
babolivier: Ah, actually, it was quoting the spec, but that comment wasn't at the right place.

client.go
```diff
@@ -59,19 +59,30 @@ func NewClientWithTimeout(timeout time.Duration) *Client {
 }
 
 type federationTripper struct {
-	transport http.RoundTripper
+	transports map[string]http.RoundTripper
```
richvdh: a comment here explaining what we are mapping from and to would be helpful

client.go
```diff
@@ -59,19 +59,30 @@ func NewClientWithTimeout(timeout time.Duration) *Client {
...
 }
 
 func newFederationTripper() *federationTripper {
-	// TODO: Verify ceritificates
-	return &federationTripper{
-		transport: &http.Transport{
+	tripper := new(federationTripper)
```
richvdh: I'm no go expert (as you will have gathered...) but is this more idiomatic than:

```
 &federationTripper{
    transports:  make(map[string]http.RoundTripper),
 }
```
?

client.go
```diff
@@ -59,19 +59,30 @@ func NewClientWithTimeout(timeout time.Duration) *Client {
...
 			TLSClientConfig: &tls.Config{
+				ServerName: tlsServerName,
 				// TODO: Remove this when we enforce MSC1711.
 				InsecureSkipVerify: true,
 			},
-		},
+		}
 	}
+
+	return f.transports[tlsServerName]
```
richvdh: can we not `return transport` here rather than doing the lookup again?

client.go
```diff
@@ -59,19 +59,30 @@ func NewClientWithTimeout(timeout time.Duration) *Client {
...
+	tripper := new(federationTripper)
+	tripper.transports = make(map[string]http.RoundTripper)
+	return tripper
+}
+
+// getTransport returns a http.Transport instance with a TLS configuration using
+// the given server name for SNI. It also creates the instance if there isn't
+// any for this server name.
+func (f *federationTripper) getTransport(tlsServerName string) http.RoundTripper {
+	if transport, ok := f.transports[tlsServerName]; !ok || transport == nil {
```
richvdh: why would transport be nil?

richvdh: could you add a comment here explaining why we need to do this rather than reuse the same transport?

richvdh: do we need to worry about locking here?

resolve.go
```diff
@@ -164,44 +167,35 @@ func resolveServer(serverName ServerName, needWellKnown bool) (results []Resolut
...
 	}
 
-	// handleNoWellKnown implements steps 4 and 5 of the algorithm (as well as
-	// 3.3 and 3.4)
-	return handleNoWellKnown(serverName, host)
+	return handleNoWellKnown(serverName, host), nil
 }
 
-func handleNoWellKnown(serverName ServerName, host string) (results []ResolutionResult, err error) {
-	dnsResults, err := LookupServer(serverName)
```
richvdh: can we get rid of LookupServer now?

resolve.go
```diff
@@ -212,9 +206,9 @@ func handleNoWellKnown(serverName ServerName, host string) (results []Resolution
 			}
 
 			results = append(results, ResolutionResult{
-				Destination: fmt.Sprintf("%s:%d", target, rec.Port),
-				Host:        serverName,
-				Name:        string(serverName),
+				Destination:   net.JoinHostPort(target, strconv.Itoa(int(rec.Port))),
```
richvdh: given that we know that `target` is not an IP literal (since such things are invalid in SRV records), can we do this the easy way?

resolve.go
```diff
@@ -164,44 +167,35 @@ func resolveServer(serverName ServerName, needWellKnown bool) (results []Resolut
...
-			}
+			// We don't want to check .well-known on the result
+			return resolveServer(result.NewAddress, false)
 		}
 	}
 
-	// handleNoWellKnown implements steps 4 and 5 of the algorithm (as well as
-	// 3.3 and 3.4)
-	return handleNoWellKnown(serverName, host)
+	return handleNoWellKnown(serverName, host), nil
```
richvdh: is there any situation in which we can get here with `serverName != host` ? I don't think so, so this could be simplified.

resolve.go
babolivier: Nope, the federation tester uses it to gather DNS results.

resolve.go
babolivier: Ah, right, good point.

resolve.go
richvdh: the federation tester needs fixing, then. We shouldn't have stuff in gomatrixserverlib which is only useful for the fed tester. (In any case, it doesn't look like the right thing for the federation tester, either.)

However, I'm happy for this to be left for a separate PR. A comment on it saying that it shouldn't be used would be useful though.

client.go
babolivier: Not sure, I did it the way I'm used to without thinking much about it, but I'm under the impression that people usually either do one or the other.
Writing this comment made me look for discussions about which one may be more idiomatic, and I saw some people mention that `&Type{...}` is usually preferred in this kind of situation, so I'll change it to use that.

client.go
babolivier: Hmm, that's from a previous attempt and I changed stuff in the meantime and forgot to remove that.

babolivier: I guess we could for performance, otherwise if we have two concurrent calls to `getTransport` the worse thing that can happen afaik would be to replace a transport with another one doing the same thing.

client.go
```diff
@@ -59,30 +60,45 @@ func NewClientWithTimeout(timeout time.Duration) *Client {
 }
 
 type federationTripper struct {
-	transports map[string]http.RoundTripper
+	// transports maps an TLS server name with an HTTP transport.
+	transports      map[string]http.RoundTripper
+	transportsMutex *sync.Mutex
```
richvdh: can this just be `sync.Mutex` so you don't need to `new` it later?

client.go
```diff
@@ -59,30 +60,45 @@ func NewClientWithTimeout(timeout time.Duration) *Client {
...
 }
 
 // getTransport returns a http.Transport instance with a TLS configuration using
 // the given server name for SNI. It also creates the instance if there isn't
 // any for this server name.
-func (f *federationTripper) getTransport(tlsServerName string) http.RoundTripper {
-	if transport, ok := f.transports[tlsServerName]; !ok || transport == nil {
-		f.transports[tlsServerName] = &http.Transport{
+// We need to use one transport per TLS server name (instead of giving our round
+// tripper a single transport) because a transport is bound to a single server
```
richvdh: yes, but why?

> because there is no way to specify the TLS ServerName on a per-connection basis.

client.go
richvdh: yeah, we'd also end up with two connection pools which would take a while to expire.

resolve.go
```diff
@@ -219,7 +219,7 @@ func handleNoWellKnown(serverName ServerName, host string) (results []Resolution
 	// record was not found
 	results = []ResolutionResult{
 		ResolutionResult{
-			Destination:   net.JoinHostPort(host, strconv.Itoa(8448)),
+			Destination:   net.JoinHostPort(string(serverName), strconv.Itoa(8448)),
```
richvdh: again, do this the easy way?

`string(serverName) + ":8448"`

richvdh:
> please can we at least have a comment on LookupServer? it's just completely broken.

I also created https://github.com/matrix-org/matrix-federation-tester/issues/63

client.go
```diff
@@ -62,22 +62,22 @@ func NewClientWithTimeout(timeout time.Duration) *Client {
...
 	transports      map[string]http.RoundTripper
-	transportsMutex *sync.Mutex
+	transportsMutex sync.Mutex
 }
 
 func newFederationTripper() *federationTripper {
 	return &federationTripper{
 		transports:      make(map[string]http.RoundTripper),
-		transportsMutex: new(sync.Mutex),
+		transportsMutex: sync.Mutex{},
```
richvdh: isn't this what happens anyway, so this is redundant? or am I confused?

babolivier: Oh yes your're right, sorry.

PR #120, opened 2019-03-18
Add federation client function for getting server version

anoadragon453:
Adds a simple function to the federation client for getting the reported server version.

### DIFF
--- a/version.go
+++ b/version.go
@@ -16,11 +16,14 @@
 package gomatrixserverlib
 
 // A Version is a struct that matches the version response from a Matrix homeserver. See
-// https://matrix.org/docs/spec/server_server/unstable.html#get-matrix-federation-v1-version
+// https://matrix.org/docs/spec/server_server/r0.1.1.html#get-matrix-federation-v1-version
 type Version struct {
-	// Name is an arbitrary string that the Matrix server uses to identify itself
-	Name string `json:"name"`
-	// Version is a string that identifies the Matrix server's version, the format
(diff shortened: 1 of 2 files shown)


version.go
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package gomatrixserverlib
+
+// A Version is a struct that matches the version response from a Matrix homeserver. See
+// https://matrix.org/docs/spec/server_server/unstable.html#get-matrix-federation-v1-version
```
richvdh: it'd be better to link to the stable version

version.go
```diff
@@ -0,0 +1,26 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package gomatrixserverlib
+
+// A Version is a struct that matches the version response from a Matrix homeserver. See
```
richvdh: according to the spec, this is returned inside a `server` property

anoadragon453:
As per Golang's default values philosophy, they'll be left as `""`.

PR #121, opened 2019-03-18
Move GetVersion to client from federation_client

anoadragon453:
Make `GetVersion` a client function rather than a federation function.

### DIFF
--- a/client.go
+++ b/client.go
@@ -365,3 +365,24 @@ func (fc *Client) DoHTTPRequest(ctx context.Context, req *http.Request) (*http.R
 
 	return resp, nil
 }
+
+// GetVersion gets the version information of a homeserver.
+// See https://matrix.org/docs/spec/server_server/r0.1.1.html#get-matrix-federation-v1-version
+func (fc *Client) GetVersion(
+	ctx context.Context, s ServerName,
+) (res Version, err error) {
+	// Construct a request for version information
+	url := url.URL{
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -365,3 +365,24 @@ func (fc *Client) DoHTTPRequest(ctx context.Context, req *http.Request) (*http.R
...
 }
+
+// GetVersion gets the version information of a homeserver.
+// See https://matrix.org/docs/spec/server_server/r0.1.1.html#get-matrix-federation-v1-version
+func (fc *Client) GetVersion(
+	ctx context.Context, s ServerName,
+) (res Version, err error) {
+	// Construct a request for version information
+	url := url.URL{
+		Scheme: "https",
```
richvdh: matrix? Check the other methods in this file.

client.go
```diff
@@ -365,3 +365,24 @@ func (fc *Client) DoHTTPRequest(ctx context.Context, req *http.Request) (*http.R
 
 	return resp, nil
 }
+
+// GetVersion gets the version information of a homeserver.
+// See https://matrix.org/docs/spec/server_server/r0.1.1.html#get-matrix-federation-v1-version
+func (fc *Client) GetVersion(
```
richvdh: this should be before the things it calls (ie, `DoRequestAndParseResponse`), and anyway conceptually it sits alongside `GetServerKeys` etc

PR #125, opened 2019-03-28
Check if a server name isn't empty before trying to validate it

### LINKED ISSUE #124: ParseAndValidateServerName panics if given an empty string
It then complains about `index out of range`.

babolivier:
Fixes #124

### DIFF
--- a/federationtypes.go
+++ b/federationtypes.go
@@ -28,7 +28,6 @@ func ParseAndValidateServerName(serverName ServerName) (host string, port int, v
 	}
 
 	host, port = splitServerName(serverName)
-	valid = false
 
 	// the host part must be one of:
 	//  - a valid (ascii) dns name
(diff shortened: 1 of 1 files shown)


federationtypes.go
```diff
@@ -22,6 +22,11 @@ type ServerName string
 //
 // if there is no explicit port, returns '-1' as the port.
 func ParseAndValidateServerName(serverName ServerName) (host string, port int, valid bool) {
+	// Don't go any further if the server name is an empty string.
+	if len(serverName) == 0 {
+		return
+	}
+
 	host, port = splitServerName(serverName)
 	valid = false
```
richvdh: seems like either this is redundant or should move up above the `if` statement

babolivier: the `valid = false` line is a bit useless here because a boolean that's not set to any value will be false. Removing it.

PR #126, opened 2019-03-29
Move FetchKeysDirect to federation tester

babolivier:
Currently this function comes with quite an unnecessary complexity, making things like https://github.com/matrix-org/matrix-federation-tester/issues/10 slightly harder to fix.

The main change is that it uses a `http.Client` to handle making the request instead of writing a `http.Request` into a TLS connection, which changes the request a bit. The federation tester is the only project in all of GitHub (including `gomatrixserverlib`) that uses this function, and this change doesn't have any impact on it (as long as its call changes to include the `timeout` parameter, which I'll do in a separate PR).

While I was at it, I also implemented support for custom timeouts.

This PR doesn't add any dependency in `*.go` files, so I assume the new entries to `go.*` are probably due to me not correctly updating them beforehand (maybe we should add a check in the CI to make sure PRs come with the correct `go.mod` and `go.sum` files).


keys.go
```diff
@@ -97,52 +95,46 @@ func (keys ServerKeys) PublicKey(keyID KeyID, atTS Timestamp) []byte {
...
-	// Create a TLS connection.
-	tcpconn, err := net.Dial("tcp", addr)
-	if err != nil {
-		return nil, nil, err
+// Optionally sets a timeout to the HTTP client if ``timeout`` isn't 0.
+// Note that this function doesn't check the validity of the certificate(s)
+// served by the server.
+// Returns the server keys and the state of the TLS connection used to retrieve
+// them.
+func FetchKeysDirect(addr, sni string, timeout time.Duration) (*ServerKeys, *tls.ConnectionState, error) {
```
richvdh: can you document the params?

keys.go
```diff
@@ -97,52 +95,46 @@ func (keys ServerKeys) PublicKey(keyID KeyID, atTS Timestamp) []byte {
...
-	})
-	if err = tlsconn.Handshake(); err != nil {
-		return nil, nil, err
-	}
-	connectionState := tlsconn.ConnectionState()
 
-	// Write a GET /_matrix/key/v2/server down the connection.
-	requestURL := "matrix://" + string(serverName) + "/_matrix/key/v2/server"
+	// Create a GET /_matrix/key/v2/server request.
+	requestURL := "https://" + addr + "/_matrix/key/v2/server"
```
richvdh: are you sure this sets the `Host` header correctly? I think you need to specify the original server name here. See how the federation client does it.

keys.go
babolivier:
```golang
// FetchKeysDirect fetches the matrix keys directly from the given address.
// Optionally sets a SNI header if ``sni`` is not empty.
// Optionally sets a timeout to the HTTP client if ``timeout`` isn't 0.
// Note that this function doesn't check the validity of the certificate(s)
// served by the server.
// Returns the server keys and the state of the TLS connection used to retrieve
// them.
```

Isn't there already enough documentation, since it documents `addr`, `sni` and `timeout`?

keys.go
babolivier: Good catch, fixed in https://github.com/matrix-org/gomatrixserverlib/pull/126/commits/fd69519046eba39779262e1f178dd292174a116f

babolivier:
> If this is only used in the fed tester, why not move it there?

Yeah, sounds sensible. I didn't do it here mainly because I wanted to keep the changes as minimal as possible but I guess it doesn't makes much of a difference. Will look into it on Monday.

keys.go
richvdh: sorry yes, I guess there is. I didn't really read it properly.

babolivier:
Paired with https://github.com/matrix-org/matrix-federation-tester/pull/78, which moves it to the federation tester.

babolivier:
> Is it just that these are indirect deps which should have been listed previously?

Yes.

> In which case, they should probably be in a separate PR?

Roger that, will do that then.

PR #129, opened 2019-05-27
Implement publicRooms API for federation client

Cnly:
Spec implemented: https://matrix.org/docs/spec/server_server/r0.1.1.html#get-matrix-federation-v1-publicrooms


anoadragon453:
Just clean up the conflict and you should be good to go.

anoadragon453:
Fixed what was breaking the tests. The end of a function had fallen off.

PR #132, opened 2019-07-25
Support auth check of redaction events in rooms >= v3

Cnly:
This also adds a `RoomVersion` field in `createContent`.

Signed-off-by: Alex Chen <minecnly@gmail.com>

### DIFF
--- a/eventauth.go
+++ b/eventauth.go
@@ -638,7 +638,7 @@ func redactEventAllowed(event Event, authEvents AuthEventProvider) error {
 
 	roomVersion := allower.create.RoomVersion
 	if roomVersion != nil && *roomVersion != "1" && *roomVersion != "2" {
-		// We always accept redaction events into DAG for rooms >= v3 after the
+		// We always accept redaction events into the DAG for rooms >= v3 after the
 		// very basic checks.
 		return nil
 	}
--- a/eventauth.go
+++ b/eventauth.go
@@ -618,7 +618,11 @@ func checkUserLevels(senderLevel int64, senderID string, oldPowerLevels, newPowe
 	return nil
 }
 
-// redactEventAllowed checks whether the m.room.redaction event is allowed.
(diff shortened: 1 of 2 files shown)


eventauth.go
```diff
@@ -632,7 +636,15 @@ func redactEventAllowed(event Event, authEvents AuthEventProvider) error {
 		return err
 	}
 
-	senderDomain, err := domainFromID(event.Sender())
+	roomVersion := allower.create.RoomVersion
+	if roomVersion != nil && *roomVersion != "1" && *roomVersion != "2" {
```
anoadragon453: I assume we should start defining these as constants at some point, but that can wait until full room version support is implemented.

eventauth.go
```diff
@@ -632,7 +636,15 @@ func redactEventAllowed(event Event, authEvents AuthEventProvider) error {
...
+		// We always accept redaction events into DAG for rooms >= v3 after the
```
anoadragon453:
```suggestion
		// We always accept redaction events into the DAG for rooms >= v3 after the
```

eventcontent.go
```diff
@@ -35,6 +35,8 @@ type createContent struct {
 	Federate *bool `json:"m.federate"`
 	// The creator of the room tells us what the default power levels are.
 	Creator string `json:"creator"`
+	// The version of the room. Should be treated as "1" when the key doesn't exist.
+	RoomVersion *string `json:"room_version"`
```
anoadragon453:
```suggestion
	RoomVersion *string `json:"room_version,omitempty"`
```

Cnly: Just a note that the `m.federate` key above is also optional. But `createContent` is not a public struct and has limited use so this should not be a big deal.

anoadragon453: I see, I suppose anything we send back to the client should have `"1"` in there anyways.

PR #137, opened 2019-08-13
Fix failing tests for the tokens package

Cnly:
Previously, these tests failed because of an unchecked error, an Errorf call
missing formatting directives, and a token duration value set wrong. They've
gone undetected because the CI runs `go test`, not `go test ./...` and thus
they've not been run.

This PR also refines some docs.

Signed-off-by: Alex Chen <minecnly@gmail.com>

### DIFF
--- a/tokens/tokens_handlers_test.go
+++ b/tokens/tokens_handlers_test.go
@@ -14,7 +14,6 @@ package tokens
 
 import (
 	"testing"
-	"time"
 )
 
 var (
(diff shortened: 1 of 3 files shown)


tokens/tokens_handlers_test.go
```diff
@@ -30,18 +29,19 @@ var (
...
-	op.Duration = 0
+	// This will set the expiry to 1 second ago
+	op.Duration = -1
 	return op
 }
 
 func TestExpiredLoginToken(t *testing.T) {
-	fakeToken, err := GenerateLoginToken(expireZeroValidTokenOp())
-	// token uses 1 second precision
-	time.Sleep(time.Second)
```
anoadragon453: Was this not important?

Cnly: I guess the intention was to generate a token that would expire "now" (reflected by `op.Duration = 0`) so a `time.Sleep` will guarantee the expiration (which won't work because `GenerateLoginToken` replaces the `0` value). Now I just `op.Duration = -1` so it's already expired "1 second ago".

P.S. Even though it expires "now", a sleep is not necessary because `ValidateToken` requires `now < expiry`.

PR #138, opened 2019-08-13
Use correct go test comamnd

anoadragon453:
The `hooks/pre-commit` script only runs `go test`, whereas it should be `go test ./...` to run tests in every package.

Relies on #137.


anoadragon453:
Succeeded by https://github.com/matrix-org/gomatrixserverlib/pull/139

PR #143, opened 2019-08-20
Expose .well-known cache expiry timestamp

reivilibre:
I'll be using this in `matrix-federation-tester` but I imagine homeservers might find it useful too.


Kegsay:
Please re-base this and I'll take another look. I'm happy to review/merge once this is done.

PR #144, opened 2019-10-01
Missing "origin" field added to RespSendJoin

behouba:
In this PR, i have added the missing `"origin"` field of `RespSendJoin` and also made some changes to `TestRespSendJoinMarshalJSON`.

### DIFF
--- a/federationtypes.go
+++ b/federationtypes.go
@@ -131,8 +131,6 @@ type RespState struct {
 	StateEvents []Event `json:"pdus"`
 	// A list of events needed to authenticate the state events.
 	AuthEvents []Event `json:"auth_chain"`
-	// The resident server's DNS name. This field is only for type compatibility with RespSendJoin and should be empty.
-	Origin string `json:"origin,omitempty"`
 }
 
 // RespPublicRooms is the content of a response to GET /_matrix/federation/v1/publicRooms
@@ -250,9 +248,13 @@ func (r RespState) Events() ([]Event, error) {
 
 // Check that a response to /state is valid.
 func (r RespState) Check(ctx context.Context, keyRing JSONVerifier) error {
+	return checkStateAndAuthChain(ctx, r.StateEvents, r.AuthEvents, keyRing)
+}
+
+func checkStateAndAuthChain(ctx context.Context, StateEvents []Event, AuthEvents []Event, keyRing JSONVerifier) error {
 	logger := util.GetLogger(ctx)
 	var allEvents []Event
-	for _, event := range r.AuthEvents {
+	for _, event := range AuthEvents {
 		if event.StateKey() == nil {
 			return fmt.Errorf("gomatrixserverlib: event %q does not have a state key", event.EventID())
 		}
@@ -260,7 +262,7 @@ func (r RespState) Check(ctx context.Context, keyRing JSONVerifier) error {
 	}
 
 	stateTuples := map[StateKeyTuple]bool{}
-	for _, event := range r.StateEvents {
+	for _, event := range StateEvents {
 		if event.StateKey() == nil {
 			return fmt.Errorf("gomatrixserverlib: event %q does not have a state key", event.EventID())
 		}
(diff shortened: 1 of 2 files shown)


Cnly:
@behouba Yes, I think that would do it.

behouba:
I have made `RespSendJoin` it own struct that do no more have it own implemantion of `json.Unmarshaler` and `json.Marshaler`. 

I have also found type casting between `RespSendJoin` and `RespState` in dendrite, so this change wil break dendrite.
@Cnly What do you think about it ?

federationtypes.go
```diff
@@ -248,17 +248,21 @@ func (r RespState) Events() ([]Event, error) {
 
 // Check that a response to /state is valid.
 func (r RespState) Check(ctx context.Context, keyRing JSONVerifier) error {
+	return checkStateAndAuthChain(ctx, r.StateEvents, r.AuthEvents, keyRing)
+}
+
+func checkStateAndAuthChain(ctx context.Context, StateEvents []Event, AuthEvents []Event, keyRing JSONVerifier) error {
```
Cnly: `StateEvents` and `AuthEvents` should be in `lowerCamelCase`.

You may also consider adding a docstring for this function.

federationtypes.go
```diff
@@ -305,51 +309,10 @@ type RespMakeJoin struct {
 
 // A RespSendJoin is the content of a response to PUT /_matrix/federation/v1/send_join/{roomID}/{eventID}
 // It has the same data as a response to /state, but in a slightly different wire format.
-type RespSendJoin RespState
-
-// MarshalJSON implements json.Marshaller
-func (r RespSendJoin) MarshalJSON() ([]byte, error) {
```
Cnly: These are still needed for now I think... They may just get removed at some point in the future.

behouba:
@Cnly has you suggested i have added a method to `func (r RespSendJoin) ToRespState() RespState` to type `RespSendJoin`. I also found this alternative much better. 
Now we should remplace each type casting from `RespSendJoin` to `RespState` with a call to `ToResState()` on the given `RespSendJoin`.

federationtypes.go
```diff
@@ -347,9 +351,18 @@ func (r *RespSendJoin) UnmarshalJSON(data []byte) error {
...
 
+// ToRespState return new RespState with the same data from the given RespSendJoin
+func (r RespSendJoin) ToRespState() RespState {
+	return RespState{
+		StateEvents: r.StateEvents,
+		AuthEvents:  r.AuthEvents,
+	}
+}
+
 type respSendJoinFields struct {
```
Cnly: Since this struct is just an adaptor between the original `RespState` and `RespSendJoin`, it can now be removed. Just move these field tags (`json:"..."`) over to the `RespSendJoin` struct and we should be good to go!

behouba: After trying to get rid of `respSendJoinFields `  i can't find a way to properly make `RespSendJoin` implement `json.MarshalJSON` and `json.UnmarshalJSON` interfaces without a type that is indirectly like `RespSendJoin`. 

 Currently the way `RespSendJoin` implement `json.MarshalJSON` and `json.UnmarshalJSON` interfaces, i think `respSendJoinFields` is still needed. 

I also found the same technique being used on `RespInvite`, i think this is because we have to deal with an array that hold data of differents types. 

However i'am still trying to find a work around.

Cnly: Oh, so there's going to be a recursive call if we remove this struct, is this what you meant? If that's the case, then I think we can leave the struct there for now. A comment on that can be added to the struct still, for clarity.

behouba: That's exactly what I meant

federationtypes.go
```diff
@@ -347,9 +351,18 @@ func (r *RespSendJoin) UnmarshalJSON(data []byte) error {
 	return nil
 }
 
+// ToRespState return new RespState with the same data from the given RespSendJoin
```
Cnly:
```suggestion
// ToRespState returns a new RespState with the same data from the given RespSendJoin
```

behouba: I will make a new commit soon.

Cnly: You can also just use GitHub's accept suggestion feature from web.

behouba: Great! 😃 I wasn't aware of this feature . I made a few additional changes (comments and tests)

behouba:
@Cnly I did not notice the existence `ServerName` type before. So i wonder if `Origin` field of `RespSendJoin` should be of type `ServerName` instead of `string`.

Cnly:
@behouba Oops... I didn't notice that during my review. 🤦 It should be `ServerName`. Could you fix that in a new PR please?

PR #145, opened 2019-10-03
Updated filters to match r0.5.0 specs

CromFr:
I just updated the filter structs to match names and orders of the [r0.5.0](https://matrix.org/docs/spec/client_server/r0.5.0.html#filtering) specs.

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -62,14 +62,16 @@ type StateFilter struct {
 
 // RoomEventFilter is used to define filtering rules for events in rooms
 type RoomEventFilter struct {
-	Limit       int      `json:"limit,omitempty"`
-	NotSenders  []string `json:"not_senders,omitempty"`
-	NotTypes    []string `json:"not_types,omitempty"`
-	Senders     []string `json:"senders,omitempty"`
-	Types       []string `json:"types,omitempty"`
-	NotRooms    []string `json:"not_rooms,omitempty"`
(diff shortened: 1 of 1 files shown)


filter.go
```diff
@@ -60,31 +83,59 @@ func (filter *Filter) Validate() error {
...
+	return EventFilter{
 		Limit:      20,
 		NotSenders: nil,
 		NotTypes:   nil,
 		Senders:    nil,
 		Types:      nil,
 	}
 }
+
+// DefaultStateFilter returns the default state event filter used by the Matrix server if no filter is provided in the request
```
babolivier: More of a nitpick, but this comment is quite long, a bit too much for a single line to my liking, could you please wrap it to a more sensible line length (80-90)?

filter.go
```diff
@@ -60,31 +83,59 @@ func (filter *Filter) Validate() error {
...
+		Types:                   nil,
+		LazyLoadMembers:         false,
+		IncludeRedundantMembers: false,
+		NotRooms:                nil,
+		Rooms:                   nil,
+		ContainsURL:             nil,
+	}
+}
+
+// DefaultRoomEventFilter returns the default room event filter used by the Matrix server if no filter is provided in the request
```
babolivier: Ditto

CromFr:
Thanks for the review !
It should be better now

PR #150, opened 2020-01-09
v2 endpoints for send_join, send_leave

neilalexander:
This changes gomatrixserverlib to use the v2 endpoints for `/send_join` and `/send_leave` and removes the expectation of needing to unmarshal `[200, ...]` from the responses. It also returns `RespSendJoin` etc to being composites of `RespState`.


go.mod
```diff
@@ -14,3 +14,5 @@ require (
 	gopkg.in/macaroon.v2 v2.1.0
 	gopkg.in/yaml.v2 v2.2.2
 )
+
+go 1.13
```
babolivier: You probably want to drop this line, otherwise people might not be able to use the lib with a lower version of Go. It's annoying that Go always adds it even when it's not needed :/

neilalexander: Good catch - fixed.

federationtypes.go
```diff
@@ -306,9 +306,8 @@ type RespMakeJoin struct {
 // A RespSendJoin is the content of a response to PUT /_matrix/federation/v1/send_join/{roomID}/{eventID}
```
richvdh: v2

federationtypes.go
```diff
@@ -317,23 +316,33 @@ func (r RespSendJoin) MarshalJSON() ([]byte, error) {
 	// of a two element list where the first element is the constant integer 200.
```
richvdh: comment is wrong

federationtypes.go
```diff
@@ -317,23 +316,33 @@ func (r RespSendJoin) MarshalJSON() ([]byte, error) {
...
+		StateEvents: r.StateEvents,
+		AuthEvents:  r.AuthEvents,
+		Origin:      r.Origin,
+	})
 }
 
 // UnmarshalJSON implements json.Unmarshaller
 func (r *RespSendJoin) UnmarshalJSON(data []byte) error {
-	var tuple []RawJSON
+	/*var tuple []RawJSON
```
richvdh: git is a thing: remove dead code, don't comment it

federationtypes_test.go
```diff
@@ -56,36 +58,33 @@ func TestRespSendJoinMarshalJSON(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	gotBytes, err := json.Marshal(RespSendJoin(input))
+	gotBytes, err := json.Marshal(input) // json.Marshal(RespSendJoin(input))
```
richvdh: dead code is dead

federationtypes.go
neilalexander: Was sure I'd removed this but apparently had reset since :-) fixed.

PR #154, opened 2020-02-28
Fix hashing algorithm

neilalexander:
This fixes the event hashing algorithm so that it correctly excludes `signatures` now. It also adds a couple of test from the spec appendies to prove that it works as intended.


Kegsay:
LGTM - the reason why this worked before is that `addContentHashesToEvent` is called before we add the signature, so there's no key to delete.

PR #156, opened 2020-03-06
Return MissingAuthEventError when auth events are missing

neilalexander:
We really should return concrete error types for more error conditions but this at least lets me fix the problem where fetching missing auth events over federation in Dendrite.


Kegsay:
LGTM - and totally agree with the sentiment of error types - the form you've done, not sentinel values ala `sql.ErrNoRows`.

PR #157, opened 2020-03-16
Define room version characteristics and types

neilalexander:
This defines the core characteristics of each room version, and the exported types that Dendrite will use to implement them.

### DIFF
--- a/eventversion.go
+++ b/eventversion.go
@@ -1,5 +1,7 @@
 package gomatrixserverlib
 
+import "fmt"
+
 // RoomVersion refers to the room version for a specific room.
 type RoomVersion string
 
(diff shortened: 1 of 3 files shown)


eventversion.go
```diff
@@ -0,0 +1,118 @@
...
+		stateResAlgorithm:      StateResV2,
+		eventFormat:            EventFormatV2,
+		eventIDFormat:          EventIDFormatV3,
+		enforceSignatureChecks: true,
+	},
+}
+
+// RoomVersionMeta contains information about a given room version, e.g. which
+// state resolution algorithm or event ID format to use.
+type roomVersion struct {
```
Kegsay: This isn't called `RoomVersionMeta`. Pick a name, any name...

neilalexander: Forgot to update the comment previously. It's fixed now.

eventversion.go
```diff
@@ -0,0 +1,118 @@
...
+	eventIDFormat          EventIDFormat
+	enforceSignatureChecks bool
+}
+
+// StateResAlgorithm returns the state resolution for the given room version.
+func (v RoomVersion) StateResAlgorithm() StateResAlgorithm {
+	if r, ok := roomVersionMeta[v]; ok {
+		return r.stateResAlgorithm
+	}
+	return StateResV1
```
Kegsay: Why do we fall back to v1 if the version is bogus? Also, you can probably make this a bit cleaner if `roomVersionMeta[""]` had the defaults in it, then you just `return r.stateResAlgorithm`.

Kegsay: Though really indexing with an invalid version should just panic.

neilalexander: Made it panic instead if the version is unrecognised.

Kegsay:
We *have* to check the room version is a version we know how to handle, otherwise fail gracefully. The current API shape doesn't make this easy, since it defaults to V1 values on bogus versions.

neilalexander:
I suppose the question is whether we should `panic()` or return an `error` for each of those functions.

neilalexander:
I've implemented errors now rather than panicking.

PR #158, opened 2020-03-16
Room version metadata and HeaderedEvent

neilalexander:
This PR implements `HeaderedEvent`, which allows bundling additional metadata like room versions along with an event and having the metadata combined/separated correctly when marshalling to/unmarshalling from JSON.

It also includes some new types and room version descriptions which are used by Dendrite.

This is the first of a series of PRs that will implement room versions.

### DIFF
--- a/event.go
+++ b/event.go
@@ -86,13 +86,6 @@ func (eb *EventBuilder) SetUnsigned(unsigned interface{}) (err error) {
 	return
 }
 
-// A VersionedEvent is a wrapper around an Event that contains information
-// about the room version.
-type VersionedEvent struct {
-	RoomVersion RoomVersion
-	Event
-}
-
 // An Event is a matrix event.
 // The event should always contain valid JSON.
 // If the event content hash is invalid then the event is redacted.
--- a/event.go
+++ b/event.go
@@ -622,6 +622,17 @@ func (e Event) MarshalJSON() ([]byte, error) {
 	return e.eventJSON, nil
 }
 
+// Headered returns a HeaderedEvent encapsulating the original event, with the
+// supplied headers.
+func (e Event) Headered(roomVersion RoomVersion) HeaderedEvent {
+	return HeaderedEvent{
+		EventHeader: EventHeader{
+			RoomVersion: roomVersion,
+		},
+		Event: e,
+	}
+}
+
 // UnmarshalJSON implements json.Unmarshaller
 func (er *EventReference) UnmarshalJSON(data []byte) error {
 	var tuple []RawJSON
--- a/event.go
+++ b/event.go
@@ -86,6 +86,13 @@ func (eb *EventBuilder) SetUnsigned(unsigned interface{}) (err error) {
 	return
 }
 
+// A VersionedEvent is a wrapper around an Event that contains information
+// about the room version.
+type VersionedEvent struct {
+	RoomVersion RoomVersion
+	Event
+}
+
 // An Event is a matrix event.
 // The event should always contain valid JSON.
 // If the event content hash is invalid then the event is redacted.
--- a/headeredevent.go
+++ b/headeredevent.go
@@ -1,0 +1,35 @@
+package gomatrixserverlib
+
+import (
+	"encoding/json"
+	"fmt"
+)
+
+// HeaderedEventHeader contains header fields for an event that contains
+// additional metadata, e.g. room version.
+type EventHeader struct {
+	RoomVersion RoomVersion `json:"room_version"`
+}
+
+// HeaderedEvent is a wrapper around an Event that contains information
+// about the room version.
+type HeaderedEvent struct {
+	EventHeader
+	Event
+}
+
+// UnmarshalJSON implements json.Unmarshaller
+func (e *HeaderedEvent) UnmarshalJSON(data []byte) error {
+	var m EventHeader
+	if err := json.Unmarshal(data, &m); err != nil {
+		return err
+	}
+	switch m.RoomVersion {
+	case RoomVersionV1, RoomVersionV2:
+		fmt.Println("room v1 or v2")
+	}
+	if err := json.Unmarshal(data, e); err != nil {
+		return err
+	}
+	return nil
+}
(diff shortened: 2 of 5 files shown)


Kegsay:
Didn't  I look at a bunch of this before?

event.go
```diff
@@ -86,6 +86,13 @@ func (eb *EventBuilder) SetUnsigned(unsigned interface{}) (err error) {
 	return
 }
 
+// A VersionedEvent is a wrapper around an Event that contains information
+// about the room version.
+type VersionedEvent struct {
```
Kegsay: Can we get rid of this? We have `HeaderedEvent`.

neilalexander: Whoops, forgot to delete this when I added `headeredevent.go`. Thanks for pointing it out.

headeredevent.go
```diff
@@ -0,0 +1,84 @@
...
+// code for marshalling and unmarshalling headered events will not work.
+type EventHeader struct {
+	RoomVersion RoomVersion `json:"room_version"`
+}
+
+// HeaderedEvent is a wrapper around an Event that contains information
+// about the room version. All header fields will be added into the event
+// when marshalling into JSON and will be separated out when unmarshalling.
+type HeaderedEvent struct {
+	EventHeader
```
Kegsay: Naming and JSON marshalling: This would read better if the field was called `Header`, rather than struct embedding like this. Also, how does this work for cases where we accidentally use the same JSON tag as an actual event value? The answer is, it panics.

This might be fine as long as we document it, as the utility of `headeredEvent.RoomVersion` may outweigh the fail-fast panic and the longness of `headeredEvent.Header.RoomVersion` everywhere.

Kegsay: https://play.golang.org/p/ijjVmbVHACe

headeredevent.go
```diff
@@ -0,0 +1,84 @@
...
+	Event
+}
+
+// UnmarshalJSON implements json.Unmarshaller
+func (e *HeaderedEvent) UnmarshalJSON(data []byte) error {
+	var err error
+	// First extract the headers from the JSON.
+	var m EventHeader
+	if err = json.Unmarshal(data, &m); err != nil {
```
Kegsay: So doing this will needlessly cause `Unmarshal` to process all fields twice: once here and once on `:53`. I was hoping to see something like:
```
struct {
    Header EventHeader `json:"header"`
    Event json.RawMessage `json:"event"`
}
```
allowing the unmarshal to skip over the rest of the event, but we don't even namespace the header to allow for this operation. What's the rationale for sharing namespaces?

headeredevent.go
neilalexander: So the main reason I chose `EventHeader` was because the entire of gomatrixserverlib only implements the `gomatrixserverlib` package, not subpackages. It makes it a bit clearer in `godoc` and friends. 

As for `struct field Bar repeats json tag` and duplicate struct tags, that's a documented Go case but I'm happy to add some more notes on it.

headeredevent.go
```diff
@@ -0,0 +1,84 @@
...
+func (e HeaderedEvent) MarshalJSON() ([]byte, error) {
+	// First marshal the event struct itself.
+	content, err := json.Marshal(e.Event)
+	if err != nil {
+		return []byte{}, err
+	}
+	// Now jump through the fields of the header struct and add them
+	// in separately. This is needed because of the way that Go handles
+	// function overloading on embedded types. Doing this ensures the
+	// least number of changes to event references elsewhere.
```
Kegsay: What does function overloading have to do with serialising fields? Why can't we just use the normal marshal impl?

headeredevent.go
Kegsay: It's worth being explicit: not everyone who contributes will be knowledgable about edge cases in Go. And you can still have `type EventHeader struct` but have the field be called `Header` you know.

headeredevent.go
neilalexander: If we do this then the struct format is different for marshalling vs unmarshalling, e.g. top-level `header` and `event` keys, which are not present currently. We could change the entire top-level `EventHeader` type but then all of the `Event` function calls throughout Dendrite also need to be updated to point to the `Event` field rather than sitting at the top level, so this meant that we could preserve as much of the current API shape as possible.

headeredevent.go
neilalexander: This comment could definitely be clearer - the problem is that `Event` also implements custom marshallers and unmarshallers because of the internal `eventJSON` field and Go tries to use those instead when the struct is top-level embedded. This was the cleanest way to avoid that.

headeredevent.go
neilalexander: Indeed, although naming the field `Header` specifically stops us from getting top-level embedding. I'll update the comments.

Kegsay: What's wrong with having the header a level further down?

neilalexander: From a struct perspective, nothing at all. From a marshalling perspective, it doesn't make things much easier for us since we still need to handle that marshalling separately or otherwise Go just defaults to the `Event` marshaller (since it overloads the top-level) and it misses out the `EventHeader` fields altogether.

headeredevent.go
Kegsay: > all of the Event function calls throughout Dendrite also need to be updated to point to the Event field rather than sitting at the top level

That seems reasonable, but it doesn't explain why we can't have a `Header` field to namespace the header/fields correctly e.g:
```
type HeaderedEvent struct {
    Header EventHeader
    Event
}
```

headeredevent.go
Kegsay: Ohh, I see now. Yuck. I still think it would be clearer to have them namespaced rather than sitting together at the same level.

headeredevent.go
neilalexander: So I've just given this a quick try and it adds more complexity into the unmarshaller, since it's then not possible to have a top-level embedded `Event` in the struct and still avoid unmarshalling it twice to separate the headers.

I think for now I'm leaning towards leaving this as it is, since in practice it's only really going to be used in a small number of cases. We could prefix the room version name tag or something, like `_room_version`, just to make it really obvious that it doesn't belong there if the headers do leak.

Kegsay: Okay, JSON prefix with `_` SGTM for now.

PR #159, opened 2020-03-24
Headered events, support for multiple event formats

neilalexander:


clientevent.go
```diff
@@ -62,3 +71,21 @@ func ToClientEvent(se Event, format EventFormat) ClientEvent {
 	}
 	return ce
 }
+
+// ToClientEvent converts a single server event to a client event.
+func HeaderedToClientEvent(se HeaderedEvent, format EventFormat) ClientEvent {
+	ce := ClientEvent{
+		Content:        RawJSON(se.Content()),
```
Kegsay: Don't we need to handle redacted content?

event.go
```diff
@@ -60,9 +62,9 @@ type EventBuilder struct {
 	// The state_key of the event if the event is a state event or nil if the event is not a state event.
 	StateKey *string `json:"state_key,omitempty"`
 	// The events that immediately preceded this event in the room history.
-	PrevEvents []EventReference `json:"prev_events"`
+	PrevEvents interface{} `json:"prev_events"`
 	// The events needed to authenticate this event.
-	AuthEvents []EventReference `json:"auth_events"`
+	AuthEvents interface{} `json:"auth_events"`
```
Kegsay: I really would prefer we didn't use `interface{}` and lose all types. Why are we doing this, there's zero justifcation or docs here and I would expect to see it.

event.go
```diff
@@ -138,12 +148,35 @@ func (eb *EventBuilder) Build(eventID string, now time.Time, origin ServerName,
...
 	}
-	if event.AuthEvents == nil {
-		event.AuthEvents = emptyEventReferenceList
+	switch eventFormat {
+	case EventFormatV1:
+		if event.PrevEvents == nil {
+			event.PrevEvents = []EventReference{}
+		}
+		if event.AuthEvents == nil {
+			event.AuthEvents = []EventReference{}
```
Kegsay: Lack of commentary.

event.go
```diff
@@ -138,12 +148,35 @@ func (eb *EventBuilder) Build(eventID string, now time.Time, origin ServerName,
...
+		switch prevEvents := event.PrevEvents.(type) {
+		case []EventReference:
+			for _, prevEvent := range prevEvents {
+				resPrevEvents = append(resPrevEvents, prevEvent.EventID)
+			}
+		}
+		switch authEvents := event.AuthEvents.(type) {
+		case []EventReference:
+			for _, authEvent := range authEvents {
+				resAuthEvents = append(resPrevEvents, authEvent.EventID)
```
Kegsay: What? Why are we appending to `resPrevEvents`?

event.go
```diff
@@ -176,8 +208,33 @@ func (eb *EventBuilder) Build(eventID string, now time.Time, origin ServerName,
...
+			return
+		}
+		fields.fixNilSlices()
+		result.fields = fields
+	case EventFormatV2:
+		if result.eventJSON, err = sjson.DeleteBytes(eventJSON, "event_id"); err != nil {
+			return
+		}
+		var fields eventFormatV2Fields
+		if err = json.Unmarshal(eventJSON, &fields); err != nil {
```
Kegsay: Why do we unmarshal the eventJSON which contains `event_id`? Shouldn't we be using `result.eventJSON`?

event.go
```diff
@@ -176,8 +208,33 @@ func (eb *EventBuilder) Build(eventID string, now time.Time, origin ServerName,
...
+			return
+		}
+		fields.EventID, err = result.generateEventID()
+		if err != nil {
+			return
```
Kegsay: This function is now too long to really be using named return args...

event.go
```diff
@@ -192,10 +249,37 @@ func (eb *EventBuilder) Build(eventID string, now time.Time, origin ServerName,
...
+	switch eventFormat {
+	case EventFormatV1:
+		fields := eventFormatV1Fields{}
+		if err = json.Unmarshal(eventJSON, &fields); err != nil {
+			return
+		}
+		fields.fixNilSlices()
+		result.fields = fields
+	case EventFormatV2:
+		if eventJSON, err = sjson.DeleteBytes(eventJSON, "event_id"); err != nil {
```
Kegsay: Same logic is in `EventBuilder.Build`, can we factor this out?

event.go
```diff
@@ -250,23 +348,56 @@ func NewEventFromUntrustedJSON(eventJSON []byte) (result Event, err error) {
...
+	case EventFormatV1:
+		result.eventJSON = eventJSON
+		var fields eventFormatV1Fields
+		if err = json.Unmarshal(eventJSON, &fields); err != nil {
+			return
+		}
+		fields.fixNilSlices()
+		result.fields = fields
+	case EventFormatV2:
+		if result.eventJSON, err = sjson.DeleteBytes(eventJSON, "event_id"); err != nil {
```
Kegsay: Again, same logic. Factor it out please.

event.go
```diff
@@ -308,9 +439,23 @@ func (e Event) SetUnsigned(unsigned interface{}) (Event, error) {
...
+	if err != nil {
+		return Event{}, err
+	}
+	if eventIDFormat == EventIDFormatV1 {
+		fields := result.fields.(eventFormatV1Fields)
+		fields.Unsigned = unsignedJSON
+		fields.fixNilSlices()
+		result.fields = fields
+	} else {
+		fields := result.fields.(eventFormatV2Fields)
```
Kegsay: It feels odd to be setting v2 without checking that the event ID format is indeed v2? No commentary to explain why this is the right thing to do. Also, you can factor out:
```
		fields.Unsigned = unsignedJSON
		fields.fixNilSlices()
		result.fields = fields
```

event.go
```diff
@@ -336,15 +481,26 @@ func (e *Event) SetUnsignedField(path string, value interface{}) error {
...
-	e.fields.Unsigned = unsigned
+	switch fields := e.fields.(type) {
+	case eventFormatV1Fields:
+		fields.Unsigned = unsigned
+		fields.fixNilSlices()
+		e.fields = fields
+	case eventFormatV2Fields:
+		fields.Unsigned = unsigned
+		fields.fixNilSlices()
+		e.fields = fields
```
Kegsay: These 2 cases are identical.... use `fallthrough`

event.go
```diff
@@ -336,15 +481,26 @@ func (e *Event) SetUnsignedField(path string, value interface{}) error {
...
+	default:
+		panic(e.invalidFieldType())
```
Kegsay: I thought you were morally opposed to panicking?

event.go
```diff
@@ -382,21 +538,37 @@ func (e Event) KeyIDs(signingName string) []KeyID {
...
 
 // StateKey returns the "state_key" of the event, or the nil if the event is not a state event.
-func (e Event) StateKey() *string {
-	return e.fields.StateKey
+func (e *Event) StateKey() *string {
+	switch fields := e.fields.(type) {
+	case eventFormatV1Fields:
+		return fields.StateKey
+	case eventFormatV2Fields:
+		return fields.StateKey
```
Kegsay: Same thing, use `fallthrough`

event.go
```diff
@@ -414,81 +586,103 @@ const (
...
+	var fields eventFields
+	switch f := e.fields.(type) {
+	case eventFormatV1Fields:
+		if f.AuthEvents == nil || f.PrevEvents == nil {
+			return errors.New("gomatrixserverlib: auth events and prev events must not be nil")
+		}
+		fields = f.eventFields
+	case eventFormatV2Fields:
+		if f.AuthEvents == nil || f.PrevEvents == nil {
+			return errors.New("gomatrixserverlib: auth events and prev events must not be nil")
```
Kegsay: Do we actually are about `nil` or do we care about zero length? No commentary to help me identify what your intent is here.

event.go
```diff
@@ -517,101 +711,239 @@ func checkID(id, kind string, sigil byte) (domain string, err error) {
...
 
 // Unsigned returns the object under the 'unsigned' key of the event.
-func (e Event) Unsigned() []byte {
-	return []byte(e.fields.Unsigned)
+func (e *Event) Unsigned() []byte {
+	switch fields := e.fields.(type) {
+	case eventFormatV1Fields:
+		return fields.Unsigned
+	case eventFormatV2Fields:
+		return fields.Unsigned
```
Kegsay: All of this boilerplate feels unnecessary. It feels like you should have an interface so you don't need to constantly type switch..

event.go
```diff
@@ -517,101 +711,239 @@ func checkID(id, kind string, sigil byte) (domain string, err error) {
...
+func (e *Event) PrevEvents() []EventReference {
+	switch fields := e.fields.(type) {
+	case eventFormatV1Fields:
+		return fields.PrevEvents
+	case eventFormatV2Fields:
+		var result []EventReference
+		for _, id := range fields.PrevEvents {
+			result = append(result, EventReference{
+				EventID:     id,
+				EventSHA256: Base64String(id[1:]),
```
Kegsay: Why `1:`?

eventcrypto.go
```diff
@@ -125,12 +127,37 @@ func referenceOfEvent(eventJSON []byte) (EventReference, error) {
...
+		switch eventIDFormat {
+		case EventIDFormatV2:
+			encoder = base64.RawStdEncoding.WithPadding(base64.NoPadding)
+		case EventIDFormatV3:
+			encoder = base64.RawURLEncoding.WithPadding(base64.NoPadding)
+		}
+		if encoder != nil {
+			eventID = fmt.Sprintf("$%s", encoder.EncodeToString(sha256Hash[:]))
+		} else {
+			return EventReference{}, errors.New("gomatrixserverlib: this should not happen")
```
Kegsay: Unhelpful. Put a `default` and return a sensible error message please.

federationtypes.go
```diff
@@ -340,6 +404,12 @@ type respSendJoinFields struct {
 
 // ToRespState returns a new RespState with the same data from the given RespSendJoin
 func (r RespSendJoin) ToRespState() RespState {
+	if len(r.StateEvents) == 0 {
+		r.StateEvents = []Event{}
+	}
+	if len(r.AuthEvents) == 0 {
+		r.AuthEvents = []Event{}
```
Kegsay: Reading this all the time is getting old. It would probably be better to have a `NewWhateverStruct(...)` which `make(...)`s them to guarantee non-nilness. When unmarshalling, I'd make a custom `UnmarshalJSON` at this point. Ultimately, you only care about this at the JSON boundary, so try to keep this cruft scoped to the JSON boundary.

It's too error prone to rely on the current method.

federationtypes_test.go
```diff
@@ -53,7 +53,8 @@ func TestParseServerName(t *testing.T) {
 
 func TestRespSendJoinMarshalJSON(t *testing.T) {
 	inputData := `{"state":[],"auth_chain":[],"origin":""}`
-	var input respSendJoinFields
+	var input RespSendJoin
+	input.RespState.roomVersion = RoomVersionV4
```
Kegsay: Why are we using v4 in the marshal test, but v1 in the unmarshal test??

headeredevent.go
```diff
@@ -26,6 +27,16 @@ type HeaderedEvent struct {
 	Event
 }
 
+// Unwrap extracts the event object from the headered event.
+func (e *HeaderedEvent) Unwrap() Event {
+	if e.RoomVersion == "" {
+		panic("gomatrixserverlib: malformed HeaderedEvent doesn't contain room version")
```
Kegsay: Again, I thought you were morally opposed to panicking?

Kegsay: If you intend to fix this later, then you need to be adding TODOs.

event.go
neilalexander: Not possible with type switches.

event.go
neilalexander: As a long-term goal I would like to not panic, but for now this is a truly impossible situation.

event.go
neilalexander: Also not possible with type switches.

event.go
neilalexander: But then we just end up with two function implementations for each set of fields for each event format, so we don't gain on lines of code - it would probably look worse.

event.go
neilalexander: I'll comment this but mostly this is because the event ID in v3 onwards is actually the event hash. `1:` excludes the `$` sigil.

event.go
neilalexander: I'll add comments here - the main reason is that `prev_events` and `auth_events` are different formats in v1/v2 vs. v3 onwards. In v3 onwards, they are no longer `[]EventReference`, but `[]string`.

event.go
neilalexander: Typo, fixed.

headeredevent.go
neilalexander: Added a TODO for now. Ideally I don't want to panic long term, but while room versions are mid-work it is easier to highlight where things go wrong.

event.go
neilalexander: Fixed.

clientevent.go
neilalexander: I don't think in this specific case, no, as if the event is redacted then it will have been updated in the database before now.

event.go
neilalexander: Fixed.

event.go
neilalexander: For now I am erring on the side of leaving it as-is since we'd still have to instantiate empty `Event` objects at each error-return.

clientevent.go
Kegsay: Databases are dendrite stuff though, not GMSL stuff? How can we be sure?

neilalexander: We can't be. It's up to Dendrite to do the right thing.

PR #161, opened 2020-03-31
Pass pointers to events rather than copying them

neilalexander:
This updates the gomatrixserverlib API to deal with pointers to events, to reduce problems where duplicate `Event` objects (when passed by value/through value receivers) would have underlying shared `interface{}` contents. It also should just reduce memory usage a little.

This also fixes a bug or two, like with redacting events and generating event IDs.

Merge this before matrix-org/dendrite#950.

### DIFF
--- a/appservice.go
+++ b/appservice.go
@@ -18,5 +18,5 @@ package gomatrixserverlib
 // ApplicationServiceTransaction is the transaction that is sent off to an
 // application service.
 type ApplicationServiceTransaction struct {
-	Events []Event `json:"events"`
+	Events []*Event `json:"events"`
 }
--- a/event.go
+++ b/event.go
@@ -313,8 +313,6 @@ func NewEventFromUntrustedJSON(eventJSON []byte, roomVersion RoomVersion) (resul
 				return
 			}
 		}
-
-		eventJSON = redactedJSON
 	}
 
 	err = result.CheckFields()
--- a/event.go
+++ b/event.go
@@ -415,30 +415,30 @@ func (e *Event) Redact() Event {
 
 // SetUnsigned sets the unsigned key of the event.
 // Returns a copy of the event with the "unsigned" key set.
-func (e *Event) SetUnsigned(unsigned interface{}) (Event, error) {
+func (e *Event) SetUnsigned(unsigned interface{}) (*Event, error) {
+	result := *e
 	var eventAsMap map[string]RawJSON
 	var err error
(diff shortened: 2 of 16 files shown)


appservice.go
```diff
@@ -18,5 +18,5 @@ package gomatrixserverlib
 // ApplicationServiceTransaction is the transaction that is sent off to an
 // application service.
 type ApplicationServiceTransaction struct {
-	Events []Event `json:"events"`
+	Events []*Event `json:"events"`
```
Kegsay: We don't need a slice of pointers here... it's already passed by reference due to the slice.

event.go
```diff
@@ -311,24 +313,22 @@ func NewEventFromUntrustedJSON(eventJSON []byte, roomVersion RoomVersion) (resul
 				return
 			}
 		}
-
-		eventJSON = redactedJSON
 	}
 
-	result.eventJSON = eventJSON
-
```
Kegsay: Why did you remove these lines?

neilalexander: It's now done at the correct time in `populateFieldsFromJSON`, e.g. after removing `event_id` in the new event format.

event.go
```diff
@@ -357,6 +358,7 @@ func (e *Event) populateFieldsFromJSON(eventJSON []byte) error {
 		if eventJSON, err = sjson.DeleteBytes(eventJSON, "event_id"); err != nil {
 			return err
 		}
+		e.eventJSON = eventJSON
```
Kegsay: Why are we setting this each time for each format and not doing it before the `switch`?

Kegsay: Oh because this version doesn't have an event ID.

headeredevent.go
```diff
@@ -24,23 +24,22 @@ type EventHeader struct {
...
 // Unwrap extracts the event object from the headered event.
-func (e *HeaderedEvent) Unwrap() Event {
+func (e *HeaderedEvent) Unwrap() *Event {
 	if e.RoomVersion == "" {
 		// TODO: Perhaps return an error here instead of panicing
 		panic("gomatrixserverlib: malformed HeaderedEvent doesn't contain room version")
 	}
-	event := e.Event
-	event.roomVersion = e.RoomVersion
-	return event
```
Kegsay: Why did you change this?

neilalexander: Because it's no longer taken by value, therefore we can just make sure that the actual event's inner room version is right and return a reference to that directly.

neilalexander:
Superseded by #240.

PR #162, opened 2020-04-01
Try to prevent nil prev_events/auth_events again

neilalexander:
This does a bit more to ensure `auth_events` and `prev_events` are never `null`, and is a part of the fix in matrix-org/dendrite#951.

### DIFF
--- a/event.go
+++ b/event.go
@@ -194,6 +194,8 @@ func (eb *EventBuilder) Build(
 				resPrevEvents = append(resPrevEvents, prevEvent.EventID)
 			}
 			event.PrevEvents = resPrevEvents
+		case nil:
+			event.PrevEvents = []string{}
 		}
 		switch authEvents := event.AuthEvents.(type) {
 		case []string:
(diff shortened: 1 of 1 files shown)


Kegsay:
A while back I mentioned:

>  It would probably be better to have a NewWhateverStruct(...) which make(...)s them to guarantee non-nilness. When unmarshalling, I'd make a custom UnmarshalJSON at this point. Ultimately, you only care about this at the JSON boundary, so try to keep this cruft scoped to the JSON boundary.
>
> It's too error prone to rely on the current method.

What's the plan with respect to this? I can only see us accumulating more of these checks otherwise. For now, LGTM.

neilalexander:
> What's the plan with respect to this? 

It's probably not a bad idea to do this, and then to just have some notes in the `godoc` for the `EventBuilder` type suggesting a builder should be used, or we could unexport the name from the package and force use of the constructor.

Kegsay:
> we could unexport the name from the package and force use of the constructor.

This won't work nicely because then any other package which relies on the type will break because you cannot do `func Foo(ev *event)`. It's pretty common to have public structs which are intended to be initialised into a valid state via `New` and not the zero-value. We can guard inside the functions for the struct against nil fields (as what `time.Time` does for `time.loc` for example) should we decide we need to.

PR #163, opened 2020-04-02
Support code for v2 invites

neilalexander:
This PR adds support code for supporting the [`/_matrix/federation/v2/invite`](https://matrix.org/docs/spec/server_server/r0.1.3#put-matrix-federation-v2-invite-roomid-eventid) endpoint.

Maybe it is better to put this stuff into `federationtypes.go` or something, or name it better? Or is it OK as-is?

### DIFF
--- a/invitev2.go
+++ b/invitev2.go
@@ -1,0 +1,72 @@
+package gomatrixserverlib
+
+// InviteV2Request and InviteV2StrippedState are defined in
+// https://matrix.org/docs/spec/server_server/r0.1.3#put-matrix-federation-v2-invite-roomid-eventid
+
+// InviteV2Request is used in a /_matrix/federation/v2/invite request.
+type InviteV2Request struct {
+	fields struct {
+		Event           Event                   `json:"event"`
+		RoomVersion     RoomVersion             `json:"room_version"`
+		InviteRoomState []InviteV2StrippedState `json:"invite_stripped_state"`
+	}
+}
+
+// Event returns the invite event.
+func (i *InviteV2Request) Event() Event {
+	return i.fields.Event
+}
+
+// RoomVersion returns the room version of the invited room.
+func (i *InviteV2Request) RoomVersion() RoomVersion {
+	return i.fields.RoomVersion
+}
+
+// InviteRoomState returns stripped state events for the room, containing
+// enough information for the client to identify the room.
+func (i *InviteV2Request) InviteRoomState() []InviteV2StrippedState {
+	return i.fields.InviteRoomState
+}
+
+// InviteV2Request is used in a /_matrix/federation/v2/invite response.
+type InviteV2Response struct {
+	fields struct {
+		Event Event `json:"event"`
+	}
+}
+
+// Event returns the invite event.
+func (i *InviteV2Response) Event() Event {
+	return i.fields.Event
+}
+
+// InviteV2StrippedState is a cut-down set of fields from room state
+// events that allow the invited server to identify the room.
+type InviteV2StrippedState struct {
+	fields struct {
+		Content  RawJSON `json:"content"`
+		StateKey *string `json:"state_key"`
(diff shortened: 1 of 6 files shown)


invitev2.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+// InviteV2Request and InviteV2StrippedState are defined in
+// https://matrix.org/docs/spec/server_server/r0.1.3#put-matrix-federation-v2-invite-roomid-eventid
+
+func NewInviteV2Request(event *HeaderedEvent, state []InviteV2StrippedState) (
+	request InviteV2Request, err error,
+) {
+	if event.RoomVersion == "" {
+		err = errors.New("gomatrixserverlib: malformed headered event")
```
Kegsay, line 17: Say what failed e.g "missing room version". For now people can infer since we just have that as data but if and when we add more fields this message becomes unhelpful.

invitev2.go
```diff
@@ -0,0 +1,125 @@
...
+	request.fields.Event = event.Unwrap()
+	return
+}
+
+type inviteV2RequestHeaders struct {
+	RoomVersion     RoomVersion             `json:"room_version"`
+	InviteRoomState []InviteV2StrippedState `json:"invite_stripped_state"`
+}
+
+// InviteV2Request is used in a /_matrix/federation/v2/invite request.
```
Kegsay, line 33: If it's the body, just say it is. If it's a mix, say so too.

invitev2.go
```diff
@@ -0,0 +1,125 @@
...
+func (i *InviteV2Request) UnmarshalJSON(data []byte) error {
+	err := json.Unmarshal(data, &i.fields.inviteV2RequestHeaders)
+	if err != nil {
+		return err
+	}
+	eventJSON := gjson.GetBytes(data, "event")
+	if !eventJSON.Exists() {
+		return errors.New("gomatrixserverlib: request doesn't contain event")
+	}
+	i.fields.Event, err = NewEventFromUntrustedJSON([]byte(eventJSON.String()), i.fields.RoomVersion)
```
Kegsay, line 56: Why do we typecast to string then `[]byte`? Can't we just use `eventJSON` as-is? Also, we don't check if `i.fields.RoomVersion` is sane.

neilalexander: Annoyingly this is because `gjson.GetBytes()` returns a `gjson.Result` rather than a `[]byte`, and seemingly none of the functions on `gjson.Result` just return a `[]byte`.

invitev2.go
neilalexander: Fixed.

invitev2_test.go
```diff
@@ -0,0 +1,51 @@
...
+	if string(j) != expected {
+		t.Fatalf("got %q, expected %q", string(j), expected)
+	}
+}
+
+func TestStrippedState(t *testing.T) {
+	output := HeaderedEvent{}
+	expected := `{"content":{"name":"test3"},"state_key":"","type":"m.room.name","sender":"@test:localhost"}`
+
+	if err := json.Unmarshal([]byte(TestHeaderedExampleEvent), &output); err != nil {
```
Kegsay, line 37: I don't really like having the input data off in another test file. It makes it more difficult than it should be to read the test. It's also non-obvious that changing the test event in one file breaks tests in another file. Just duplicate the event JSON, it's clearer.

invitev2.go
neilalexander: Not sure I understand what you mean here?

Kegsay: This struct is used in a request. Where in the request? Query params? Body? Headers?

neilalexander: Ah OK, in the body it is - fixed.

invitev2_test.go
neilalexander: Fixed.

PR #167, opened 2020-04-16
Parallelise direct key fetcher

neilalexander:
This spawns goroutines per server for the direct key fetcher and waits for results to return using a wait group. The result is that fetching keys in big rooms now doesn't take forever (a maximum of 30 seconds since that appears to be our outbound timeout) and this reduces the chance that the requestor context expires before we finish:
```
Apr 16 15:41:58 photon-machine docker-compose[3829]: dendrite_monolith | time="2020-04-16T15:41:58.284643563Z" level=info msg="Checking event signatures for 7677 events of room state" func="Check\n\t" file=" [/go/pkg/mod/github.com/matrix-org/gomatrixserverlib@v0.0.0-20200416153646-47a0f63244a5/federationtypes.go:332]" req.id=4sPgj4M4LH8x req.method=POST req.path="/_matrix/client/r0/join/!BAXLHOFjvDKUeLafmO:matrix.org" user_id="@neilalexander:dendrite.neilalexander.dev"
Apr 16 15:42:00 photon-machine docker-compose[3829]: dendrite_monolith | time="2020-04-16T15:42:00.434945429Z" level=info msg="Requesting keys from fetcher" func="func1\n\t" file=" [/go/pkg/mod/github.com/matrix-org/gomatrixserverlib@v0.0.0-20200416153646-47a0f63244a5/keyring.go:196]" fetcher=DirectKeyFetcher num_key_requests=194 req.id=4sPgj4M4LH8x req.method=POST req.path="/_matrix/client/r0/join/!BAXLHOFjvDKUeLafmO:matrix.org" user_id="@neilalexander:dendrite.neilalexander.dev"
...
Apr 16 15:42:30 photon-machine docker-compose[3829]: dendrite_monolith | time="2020-04-16T15:42:30.637662978Z" level=info msg="Got keys from fetcher" func="func1\n\t" file=" [/go/pkg/mod/github.com/matrix-org/gomatrixserverlib@v0.0.0-20200416153646-47a0f63244a5/keyring.go:204]" fetcher=DirectKeyFetcher num_keys_fetched=137 req.id=4sPgj4M4LH8x req.method=POST req.path="/_matrix/client/r0/join/!BAXLHOFjvDKUeLafmO:matrix.org" user_id="@neilalexander:dendrite.neilalexander.dev"
```

### DIFF
--- a/keyring.go
+++ b/keyring.go
@@ -396,14 +396,14 @@ func (d *DirectKeyFetcher) FetchKeys(
 	}
 
 	// Define our worker.
-	worker := func(server ServerName) {
+	worker := func() {
 		defer wait.Done()
 		for {
 			if len(pending) == 0 {
 				// There are no more jobs left, so stop the worker.
 				return
 			}
-			serverResults, err := d.fetchKeysForServer(ctx, server)
+			serverResults, err := d.fetchKeysForServer(ctx, <-pending)
 			if err != nil {
 				// TODO: Should we actually be erroring here? or should we just drop those keys from the result map?
 				return
@@ -418,12 +418,13 @@ func (d *DirectKeyFetcher) FetchKeys(
 
 	// Start the workers.
 	for i := 0; i < numWorkers; i++ {
-		go worker(<-pending)
+		go worker()
 	}
 
 	// Wait for the workers to finish before returning
 	// the results.
 	wait.Wait()
+	close(pending)
 	return results, nil
 }
 
--- a/keyring.go
+++ b/keyring.go
@@ -372,14 +372,37 @@ func (d *DirectKeyFetcher) FetchKeys(
 		server[req] = ts
 	}
 
+	// Work out the number of workers that we want to start. If the
+	// number of outstanding requests is less than the current max
+	// then reduce it so we don't start workers unnecessarily.
+	numWorkers := 64
+	if len(byServer) < numWorkers {
+		numWorkers = len(byServer)
+	}
+
+	// Prepare somewhere to put the results. This map is protected
+	// by the below mutex.
 	results := map[PublicKeyLookupRequest]PublicKeyLookupResult{}
 	var resultsMutex sync.Mutex
-	var wait sync.WaitGroup
-	wait.Add(len(byServer))
 
-	for s := range byServer {
-		go func(server ServerName) {
-			defer wait.Done()
+	// Populate the wait group with the number of workers.
+	var wait sync.WaitGroup
+	wait.Add(numWorkers)
+
+	// Populate the jobs queue.
+	pending := make(chan ServerName, len(byServer))
+	for serverName := range byServer {
+		pending <- serverName
+	}
+
+	// Define our worker.
+	worker := func(server ServerName) {
+		defer wait.Done()
+		for {
+			if len(pending) == 0 {
+				// There are no more jobs left, so stop the worker.
+				return
+			}
 			serverResults, err := d.fetchKeysForServer(ctx, server)
 			if err != nil {
 				// TODO: Should we actually be erroring here? or should we just drop those keys from the result map?
@@ -390,11 +413,17 @@ func (d *DirectKeyFetcher) FetchKeys(
 			for req, keys := range serverResults {
 				results[req] = keys
 			}
-		}(s)
+		}
 	}
 
+	// Start the workers.
+	for i := 0; i < numWorkers; i++ {
+		go worker(<-pending)
+	}
+
+	// Wait for the workers to finish before returning
+	// the results.
 	wait.Wait()
-	resultsMutex.Lock()
 	return results, nil
 }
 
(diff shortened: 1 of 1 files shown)


keyring.go
```diff
@@ -372,17 +373,28 @@ func (d *DirectKeyFetcher) FetchKeys(
...
+			resultsMutex.Lock()
+			defer resultsMutex.Unlock()
+			for req, keys := range serverResults {
+				results[req] = keys
+			}
+		}(s)
 	}
+
+	wait.Wait()
+	resultsMutex.Lock()
```
Kegsay, line 397: Why do we lock here?

keyring.go
```diff
@@ -372,17 +373,28 @@ func (d *DirectKeyFetcher) FetchKeys(
...
-		if err != nil {
-			// TODO: Should we actually be erroring here? or should we just drop those keys from the result map?
-			return nil, err
-		}
-		for req, keys := range serverResults {
-			results[req] = keys
-		}
+	var resultsMutex sync.Mutex
+	var wait sync.WaitGroup
+	wait.Add(len(byServer))
```
Kegsay, line 378: If we have many servers this will be a problem. Spawning hundreds of parallel HTTP requests might hit fdlimits or cause other nasty behaviour which we won't see until we hit that threshold. A worker pool would be a better solution.

keyring.go
```diff
@@ -372,17 +373,28 @@ func (d *DirectKeyFetcher) FetchKeys(
...
+
+	for s := range byServer {
+		go func(server ServerName) {
+			defer wait.Done()
+			serverResults, err := d.fetchKeysForServer(ctx, server)
+			if err != nil {
+				// TODO: Should we actually be erroring here? or should we just drop those keys from the result map?
```
Kegsay, line 385: We can at least log it?

keyring.go
Kegsay: That doesn't have to happen now, but a TODO would be good.

keyring.go
```diff
@@ -371,18 +372,58 @@ func (d *DirectKeyFetcher) FetchKeys(
...
+			defer resultsMutex.Unlock()
+			for req, keys := range serverResults {
+				results[req] = keys
+			}
 		}
 	}
+
+	// Start the workers.
+	for i := 0; i < numWorkers; i++ {
+		go worker(<-pending)
```
Kegsay, line 421: This is weird - why bother using a channel at all? You're just sending servers down it on the same goroutine as you then read from.

keyring.go
```diff
@@ -371,18 +372,58 @@ func (d *DirectKeyFetcher) FetchKeys(
...
+	pending := make(chan ServerName, len(byServer))
+	for serverName := range byServer {
+		pending <- serverName
+	}
+
+	// Define our worker.
+	worker := func(server ServerName) {
+		defer wait.Done()
+		for {
+			if len(pending) == 0 {
```
Kegsay, line 402: We don't ever read from `pending` again so won't this loop forever in scenarios were the number of servers > workers?

keyring.go
neilalexander: Mostly because they're thread-safe by default.

keyring.go
```diff
@@ -371,18 +372,58 @@ func (d *DirectKeyFetcher) FetchKeys(
...
-		for req, keys := range serverResults {
-			results[req] = keys
+	var resultsMutex sync.Mutex
+
+	// Populate the wait group with the number of workers.
+	var wait sync.WaitGroup
+	wait.Add(numWorkers)
+
+	// Populate the jobs queue.
+	pending := make(chan ServerName, len(byServer))
```
Kegsay, line 393: This channel is never closed so it'll leak resources.

keyring.go
neilalexander: Yes, you're right, that channel read needs to be in the loop, not outwith it.

keyring.go
neilalexander: Fixed.

keyring.go
neilalexander: Should be fixed now.

keyring.go
neilalexander: It's gone now - was a mistake.

keyring.go
```diff
@@ -371,18 +372,55 @@ func (d *DirectKeyFetcher) FetchKeys(
...
+	close(pending)
+
+	// Define our worker.
+	worker := func(ch <-chan ServerName) {
+		defer wait.Done()
+		for server := range ch {
+			serverResults, err := d.fetchKeysForServer(ctx, server)
+			if err != nil {
+				// TODO: Should we actually be erroring here? or should we just drop those keys from the result map?
+				return
```
Kegsay, line 406: You don't want to return here else if the number of errors > the number of workers we will terminate all workers before going through all the requests. `continue` instead

neilalexander: Ah, yes, good catch. Missed that.

PR #168, opened 2020-04-17
Update validity check

neilalexander:
This signature checking so that strict validity checking can be optional (as is needed for room versions 1-4) whilst maintaining it for federation HTTP requests and allowing us to use it for room version 5.

It also fixes some potential panics in tests.

### DIFF
--- a/keyring_test.go
+++ b/keyring_test.go
@@ -88,9 +88,10 @@ func TestVerifyJSONsSuccess(t *testing.T) {
 	// Check that trying to verify the server key JSON works.
 	k := KeyRing{nil, &testKeyDatabase{}}
 	results, err := k.VerifyJSONs(context.Background(), []VerifyJSONRequest{{
-		ServerName: "localhost:8800",
-		Message:    []byte(testKeys),
-		AtTS:       1493142432964,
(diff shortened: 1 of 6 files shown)


neilalexander:
There are tests already and they pass (I broke them briefly by `<` instead of `<=` on the first check so I know they still work).

neilalexander:
Closing this as I think I've possibly unravelled something else.

keyring_test.go
```diff
@@ -137,9 +140,10 @@ func TestVerifyJSONsFetcherError(t *testing.T) {
 	// Check that if the database errors then the attempt to verify JSON fails.
 	k := KeyRing{nil, &erroringKeyDatabase{}}
 	results, err := k.VerifyJSONs(context.Background(), []VerifyJSONRequest{{
-		ServerName: "localhost:8800",
-		Message:    []byte(testKeys),
-		AtTS:       1493142432964,
+		ServerName:             "localhost:8800",
+		Message:                []byte(testKeys),
+		AtTS:                   1493142432964,
+		StrictValidityChecking: true,
```
Kegsay, line 146: Can you add a test where this is `false` please.

PR #169, opened 2020-04-17
SendRespJoin checks

neilalexander:
This PR primarily does two things:

- It ensures that the auth and state events are handled properly for `SendRespJoin`, and ensuring that we're checking if the event is allowed against the correct set of auth events.
- It increases `DefaultStateFilter()` to return `math.MaxUint32` limits rather than 20, since this stopped us from ever seeing full room state.

### DIFF
--- a/federationtypes.go
+++ b/federationtypes.go
@@ -238,7 +238,67 @@ func (r RespState) Events() ([]Event, error) {
 	if len(r.AuthEvents) == 0 {
 		r.AuthEvents = []Event{}
 	}
-	return ReverseTopologicalOrdering(append(r.StateEvents, r.AuthEvents...)), nil
+	eventsByID := map[string]*Event{}
+	// Collect a map of event reference to event
+	for i := range r.StateEvents {
+		eventsByID[r.StateEvents[i].EventID()] = &r.StateEvents[i]
+	}
+	for i := range r.AuthEvents {
+		eventsByID[r.AuthEvents[i].EventID()] = &r.AuthEvents[i]
+	}
+
+	queued := map[*Event]bool{}
+	outputted := map[*Event]bool{}
+	var result []Event
+	for _, event := range eventsByID {
+		if outputted[event] {
+			// If we've already written the event then we can skip it.
+			continue
+		}
+
+		// The code below does a depth first scan through the auth events
+		// looking for events that can be appended to the output.
+
+		// We use an explicit stack rather than using recursion so
+		// that we can check we aren't creating cycles.
+		stack := []*Event{event}
+
+	LoopProcessTopOfStack:
+		for len(stack) > 0 {
+			top := stack[len(stack)-1]
+			// Check if we can output the top of the stack.
+			// We can output it if we have outputted all of its auth_events.
+			for _, ref := range top.AuthEvents() {
+				authEvent := eventsByID[ref.EventID]
+				if authEvent == nil {
+					return nil, MissingAuthEventError{ref.EventID, top.EventID()}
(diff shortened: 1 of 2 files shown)


federationtypes.go
```diff
@@ -238,67 +238,7 @@ func (r RespState) Events() ([]Event, error) {
...
-			// the stack then we can append it to the result and try processing
-			// the item below it in the stack.
-			result = append(result, *top)
-			outputted[top] = true
-			stack = stack[:len(stack)-1]
-		}
-	}
-
-	return result, nil
+	return ReverseTopologicalOrdering(append(r.StateEvents, r.AuthEvents...)), nil
```
Kegsay, line 241: I don't see how this function is equivalent to the removed code. For starters, the removed code could fail due to missing auth events or event cycles, yet here this is now impossible.

federationtypes.go
```diff
@@ -429,27 +369,34 @@ func (r RespSendJoin) Check(ctx context.Context, keyRing JSONVerifier, joinEvent
...
+	stateEvents := NewAuthEvents(nil)
+
+	for i, event := range r.AuthEvents {
+		authEventsByID[event.EventID()] = &r.AuthEvents[i]
+	}
+
 	for i, event := range r.StateEvents {
-		stateEventsByID[event.EventID()] = &r.StateEvents[i]
-		if err := authEvents.AddEvent(&r.StateEvents[i]); err != nil {
+		authEventsByID[event.EventID()] = &r.StateEvents[i]
```
Kegsay, line 380: Why are we assigning to `authEventsByID` here? This doesn't make sense to me.

federationtypes.go
```diff
@@ -429,27 +369,34 @@ func (r RespSendJoin) Check(ctx context.Context, keyRing JSONVerifier, joinEvent
 		return err
 	}
 
-	stateEventsByID := map[string]*Event{}
-	authEvents := NewAuthEvents(nil)
+	authEventsByID := map[string]*Event{}
+	stateEvents := NewAuthEvents(nil)
```
Kegsay, line 373: This is a confusing name surely.

federationtypes.go
```diff
@@ -429,27 +429,42 @@ func (r RespSendJoin) Check(ctx context.Context, keyRing JSONVerifier, joinEvent
...
+	for i, event := range r.AuthEvents {
+		eventsByID[event.EventID()] = &r.AuthEvents[i]
+	}
+
+	// Then we add the current state events too, since our newly formed
+	// membership event will likely refer to these as auth events too.
 	for i, event := range r.StateEvents {
-		stateEventsByID[event.EventID()] = &r.StateEvents[i]
-		if err := authEvents.AddEvent(&r.StateEvents[i]); err != nil {
+		eventsByID[event.EventID()] = &r.StateEvents[i]
```
Kegsay, line 445: Isn't there a risk that this will set the same event twice? I suppose that's probably not a problem?

federationtypes.go
Kegsay: Why rename this? `authEventProvider` would be clearer given comments on `:448`.

federationtypes.go
neilalexander: In room versions 1 and 2 it is possible and that might break the join, but that's somewhat unavoidable. In room versions 3 and onwards, it is no longer possible because by changing the event—and by extension the reference hash—we inevitably change the event ID.

federationtypes.go
neilalexander: I've named it `authEventProvider`.

PR #170, opened 2020-04-20
Make sure we're trying next key fetchers if needed

neilalexander:
This improves multi-key-fetcher configurations so that we will continue to ask all of the fetchers for keys that we haven't received back.

### DIFF
--- a/keyring.go
+++ b/keyring.go
@@ -172,7 +172,7 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest) 
 		delete(keyRequests, req)
 	}
 
-	k.checkUsingKeys(requests, results, keyIDs, keysFromDatabase)
+	k.checkUsingKeys(requests, results, keyIDs, keysFetched)
 
 	// If we can verify using the keys from the database, don't make a federation call.
 	doFederationHit := false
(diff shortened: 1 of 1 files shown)


keyring.go
```diff
@@ -196,12 +196,25 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest)
...
 
+		// For the keys that we successfully fetched, don't try again
+		for resp := range keysFetched {
+			delete(keyRequests, resp)
+		}
+
+		// If we still have outstanding requests that weren't satisfied
+		// then try them with the next key fetcher
+		if len(keyRequests) > 0 {
+			continue
```
Kegsay, line 215: If we do have other outstanding requests, we drop the keys fetched `keysFetched` to the floor by the looks of it? Don't we need to accumulate them somewhere?

neilalexander: Good spot, yes. Fixed.

keyring.go
```diff
@@ -163,6 +163,15 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest)
 	if err != nil {
 		return nil, err
 	}
+
+	// TODO: we should distinguish here between expired keys, and those we don't have.
+	// If the key has expired, it's no use re-requesting it.
```
Kegsay, line 168: Random Q: if the key is supplied by another server (eg perspective or key query API) does the original server sign when the key expires? Else you could get a form of censorship if the intermediary server changes the expiration time maliciously.

I know the federation API signs point to point requests, but won't that be stripped away by the intermediary server? E.g A <---> B <---> C is all signed, but A <-----> C is not

PR #171, opened 2020-04-21
Backfill: Add support for making and verifying backfill requests

Kegsay:
This PR adds the function `RequestBackfill`. Comments and tests should explain everything.

Currently this does not do event auth checks as it would make this PR too big. They will come next. Some of the groundwork has been laid already in `BackfillRequester`.

The overall design here is to take away as much of the Matrix logic as possible, and just leave dumb obvious stuff for implementors to implement in `BackfillRequester`. For Dendrite, we'll use a `FederationClient` for most of these, and query the roomserver for  `ServersAtEvent`.

I've left in some commented out code to illustrate what I plan to do on the flipside: implementing the receiving end of backfill, which will follow the same pattern.


backfill.go
```diff
@@ -0,0 +1,123 @@
...
+		return nil, nil
+	}
+	haveEventIDs := make(map[string]bool)
+	var result []HeaderedEvent
+	// pick a server to backfill from
+	// TODO: use other event IDs and make a set out of all the returned servers?
+	servers := b.ServersAtEvent(ctx, roomID, fromEventIDs[0])
+	// loop each server asking it for `limit` events. Worst case, we ask every server for `limit`
+	// events before giving up. Best case, we just ask one.
+	for _, s := range servers {
```
neilalexander, line 53: This is something we're generally quite bad at throughout, but it just stood out to me here, that wherever we have HTTP requests in loops where we `continue` on error, we really should check that the initial context hasn't expired at the beginning of each iteration to save us from unnecessarily calling `Backfill` a bunch more times with an expired context. It's basically a no-op if we do but they are no-ops that are noisy in the logs.

Kegsay: Good point, done!

PR #172, opened 2020-04-21
LookupMissingEvents

neilalexander:
This adds `MissingEvents`, `RespMissingEvents` and `LookupMissingEvents()`.

### DIFF
--- a/federationclient.go
+++ b/federationclient.go
@@ -238,6 +238,7 @@ func (ac *FederationClient) LookupStateIDs(
 
 // LookupMissingEvents asks a remote server for missing events within a
 // given bracket.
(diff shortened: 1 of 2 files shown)


federationtypes.go
```diff
@@ -135,6 +135,26 @@ type RespState struct {
 	AuthEvents []Event `json:"auth_chain"`
 }
 
+// MissingEvents represents a set of missing events.
```
Kegsay, line 138: It's a request for missing events actually isn't it?

federationclient.go
```diff
@@ -236,6 +236,23 @@ func (ac *FederationClient) LookupStateIDs(
 	return
 }
 
+// LookupMissingEvents asks a remote server for missing events within a
+// given bracket.
```
Kegsay, line 240: Please link to the spec

PR #173, opened 2020-04-22
Add VerifyEventAuthChain and use it when backfilling

Kegsay:
This currently doesn't make network calls, the next PR will add an implementation of `AuthChainProvider` which will rely entirely on network calls. The intention then is for Dendrite to use that provider as a fallback, after initially trying the database/caches/querying roomserver/magic.

### DIFF
--- a/authchain.go
+++ b/authchain.go
@@ -1,0 +1,62 @@
+package gomatrixserverlib
+
+import (
+	"context"
+	"fmt"
+)
+
+// AuthChainProvider returns the requested list of auth events.
+type AuthChainProvider func(roomVer RoomVersion, eventIDs []string) ([]Event, error)
+
+// VerifyEventAuthChain will verify that the event is allowed according to its auth_events, and then
+// recursively verify each of those auth_events.
+//
+// This implements Step 4 of https://matrix.org/docs/spec/server_server/latest#checks-performed-on-receipt-of-a-pdu
+// "Passes authorization rules based on the event's auth events, otherwise it is rejected."
+//
+// The `provideEvents` function will only be called for *new* events rather than for everything as it is
+// assumed that this function is costly. Failing to provide all the requested events will fail this function.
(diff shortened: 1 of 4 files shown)


authchain.go
```diff
@@ -0,0 +1,62 @@
...
+		// fetch the events and add them to the lookup table
+		if len(need) > 0 {
+			newEvents, err := provideEvents(eventToVerify.roomVersion, need)
+			if err != nil {
+				return fmt.Errorf("gomatrixserverlib: VerifyEventAuthChain failed to obtain auth events: %w", err)
+			}
+			for i := range newEvents {
+				eventsByID[newEvents[i].EventID()] = &newEvents[i] // add to lookup table
+			}
+			eventsToVerify = append(eventsToVerify, newEvents...) // verify these events too
```
neilalexander, line 52: Might not be apparent until we see it in real use but I suspect the order of `newEvents` is important here, but I don't see anything that enforces it. On such a small set of events, the `ReverseTopologicalOrdering` function might just be enough to guarantee that you always verify possibly new prerequisite events first.

Kegsay: The order shouldn't matter, because we cache results at every stage. Each stage can independently verify a single event, so you shouldn't get any weird dependencies on requiring X to be verified *before* you verify Y. We don't need to make sure that the create event is acceptable before we verify a message event, we can do them both independently and then when the create event fails it will fail the overall request.

This does mean that it's possible to parallelise this algorithm pretty easily should the auth check itself become complex, but I don't want to prematurely optimise this for now.

neilalexander: Fair enough!

Kegsay: You could of course do some funky "well the create event is bad, so anything which references it can be automatically discarded without doing an auth check" stuff, but that's an optimisation for later. That *would* create a dependency between events.

PR #177, opened 2020-04-24
Implement step 5 of PDU checks for backfilling

Kegsay:
Technically it will also do step 6 but that needs the current forward extremity event ID.

### DIFF
--- a/authstate.go
+++ b/authstate.go
@@ -10,7 +10,7 @@ import (
 // StateProvider is capable of returning the room state at any point in time.
 type StateProvider interface {
 	// StateIDsAtEvent returns a list of state event IDs for the event ID provided, which represent the entire
-	// room state at that event. An empty `atEventID` means "the current room state".
+	// room state at that event.
 	StateIDsAtEvent(ctx context.Context, roomID, atEventID string) ([]string, error)
 	// StateAtEvent returns the state of the room at the given event. `eventIDs` will be populated with the output
 	// of StateIDsAtEvent to aid in event retrieval.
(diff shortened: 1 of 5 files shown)


authstate.go
```diff
@@ -0,0 +1,117 @@
...
+	"fmt"
+
+	"github.com/matrix-org/util"
+)
+
+// StateProvider is capable of returning the room state at any point in time.
+type StateProvider interface {
+	// StateIDsAtEvent returns a list of state event IDs for the event ID provided, which represent the entire
+	// room state at that event.
+	StateIDsAtEvent(ctx context.Context, roomID, atEventID string) ([]string, error)
```
neilalexander, line 14: `At` is an unclear term here when referring to state events. It would be better to use `Before` or `After` (where `Before` doesn't include state changes from the event and `After` does include them).

load.go
```diff
@@ -15,17 +15,24 @@ type EventLoadResult struct {
 
 // EventsLoader loads untrusted events and verifies them.
 type EventsLoader struct {
-	ver      RoomVersion
-	keyRing  JSONVerifier
-	provider AuthChainProvider
+	ver           RoomVersion
```
neilalexander, line 18: Pedantic but `roomVersion` would be nicer than `ver` for consistency/clarity.

authstate.go
Kegsay: I agree, and actually don't know the answer as the spec just says:

> Passes authorization rules based on the state at the event, otherwise it is rejected.

Kegsay: I guess `Before`, else if you say changed power levels to give yourself admin then you'd be able to as the state at that point would already be giving yourself admin.

PR #178, opened 2020-04-27
breaking: Modify the API for ReverseTopologicalOrdering and LoadAndVerify

Kegsay:
They now both require an `Ordering` which is usually `OrderingAuthEvents`
except in the case of backfilling, where we use `OrderingPrevEvents`.
We need to sort by prev_events for backfilling so we can make optimisations
and not call `/state_ids` repeatedly, as well as work out the state at
that event for snapshots/PDU checks.

### DIFF
--- a/stateresolutionv2.go
+++ b/stateresolutionv2.go
@@ -17,8 +17,18 @@ package gomatrixserverlib
 import (
 	"container/heap"
 	"encoding/json"
+	"fmt"
 	"sort"
 	"strconv"
+)
+
+// Ordering represents how to sort a list of events, used primarily in ReverseTopologicalOrdering
+type Ordering int
+
+// Sort events by prev_events or auth_events
+const (
+	OrderingPrevEvents Ordering = iota + 1
+	OrderingAuthEvents
 )
 
 type stateResolverV2 struct {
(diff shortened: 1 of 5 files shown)


stateresolutionv2.go
```diff
@@ -583,3 +603,98 @@ resetNoIncoming:
 	//sort.Sort(sort.Reverse(stateResV2ConflictedPowerLevelHeap(graph)))
 	return graph
 }
+
+// kahnsAlgorithmByAuthEvents is, predictably, an implementation of Kahn's
+// algorithm that uses auth events to topologically sort the input list of
```
neilalexander, line 608: Comments still mention "auth events here".

(There's a few instances in the comments in the function, too.)

stateresolutionv2.go
```diff
@@ -17,10 +17,20 @@ package gomatrixserverlib
 import (
 	"container/heap"
 	"encoding/json"
+	"fmt"
 	"sort"
 	"strconv"
 )
 
+// Ordering represents how to sort a list of events, used primarily in ReverseTopologicalOrdering
+type Ordering int
```
neilalexander, line 26: In the exported interface this might be a bit of a generic name, might be better as something like `TopologicalOrder`.

stateresolutionv2.go
```diff
@@ -17,10 +17,20 @@ package gomatrixserverlib
...
+
+// Sort events by prev_events or auth_events
+const (
+	OrderingPrevEvents Ordering = iota + 1
```
neilalexander, line 30: Same here—maybe `TopologicalOrderByPrevEvents` etc.

PR #181, opened 2020-04-27
Use query escaping for event ID fields in URLs

neilalexander:
Because room version 3 event IDs aren't URL-safe, we need to use query escaping rather than path escaping to stop `+`, `:` and friends from creeping in and mangling the URLs.

### DIFF
--- a/federationclient.go
+++ b/federationclient.go
@@ -96,8 +96,8 @@ func (ac *FederationClient) MakeJoin(
 		versionQueryString = "?" + strings.Join(vqs, "&")
 	}
 	path := federationPathPrefixV1 + "/make_join/" +
-		url.PathEscape(roomID) + "/" +
-		url.PathEscape(userID) + versionQueryString
+		url.QueryEscape(roomID) + "/" +
+		url.QueryEscape(userID) + versionQueryString
 	req := NewFederationRequest("GET", s, path)
 	err = ac.doRequest(ctx, req, &res)
 	return
@@ -112,7 +112,7 @@ func (ac *FederationClient) SendJoin(
 ) (res RespSendJoin, err error) {
 	res.RespState.roomVersion = roomVersion
 	path := federationPathPrefixV2 + "/send_join/" +
-		url.PathEscape(event.RoomID()) + "/" +
+		url.QueryEscape(event.RoomID()) + "/" +
 		url.QueryEscape(event.EventID())
 	req := NewFederationRequest("PUT", s, path)
 	if err = req.SetContent(event); err != nil {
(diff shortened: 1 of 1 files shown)


Kegsay:
I thought the whole point of `PathEscape` was to escape things like `/`?

Kegsay:
>  PathEscape escapes the string so it can be safely placed inside a URL path segment, replacing special characters (including /) with %XX sequences as needed.

neilalexander:
Indeed you're right that it's not the `/`, but apparently `:` and `+` → https://play.golang.org/p/jrOj8hElZoV

Using this commit, the dendrite test passes, so I guess it is unhappy with those naked characters.

Kegsay:
We need to be doing whatever makes Synapse happy too though, as if Synapse doesn't decode them the same way this won't work. What does Synapse do for this?

Kegsay:
I'd like to figure this out a bit more first, since I believe `PathEscape` works just fine, meaning there must be something else at play here.

neilalexander:
In response to what Synapse does, it also query-encodes the event ID:
```
Apr 28 08:49:28 photon-machine docker-compose[725]: synapse_1   | 2020-04-28 08:49:28,248 - synapse.federation.transport.client - 314 - DEBUG - POST-62041 - Invoked 'send_invite_v2' with args: self=<synapse.federation.transport.client.TransportLaye..., destination=matrix.org, room_id=!omICAzeOwDPIYDwtuO:synapse.neilalexander.dev, event_id=$igkK3rETJY+Bvxet/ZaNTCgowzQeeDK5HFV2AsM7q40, content={'event': {'auth_events': ['$RJ52XBm27HsNVX1iaWnMo...
Apr 28 08:49:28 photon-machine docker-compose[725]: synapse_1   | 2020-04-28 08:49:28,249 - synapse.http.matrixfederationclient - 408 - INFO - POST-62041 - {PUT-O-10708} [matrix.org] Sending request: PUT matrix://matrix.org/_matrix/federation/v2/invite/%21omICAzeOwDPIYDwtuO%3Asynapse.neilalexander.dev/%24igkK3rETJY%2BBvxet%2FZaNTCgowzQeeDK5HFV2AsM7q40; timeout 60.000000s
```

neilalexander:
OK, so according to [RFC1738](https://www.ietf.org/rfc/rfc1738.txt), the following characters are reserved:
```
reserved       = ";" | "/" | "?" | ":" | "@" | "&" | "="
```
It states:
```
Many URL schemes reserve certain characters for a special meaning:
their appearance in the scheme-specific part of the URL has a
designated semantics. If the character corresponding to an octet is
reserved in a scheme, the octet must be encoded.  The characters ";",
"/", "?", ":", "@", "=" and "&" are the characters which may be
reserved for special meaning within a scheme. No other characters may
be reserved within a scheme.

Usually a URL has the same interpretation when an octet is
represented by a character and when it encoded. However, this is not
true for reserved characters: encoding a character reserved for a
particular scheme may change the semantics of a URL.

Thus, only alphanumerics, the special characters "$-_.+!*'(),", and
reserved characters used for their reserved purposes may be used
unencoded within a URL.
```
So the answer is that we actually should always be query-encoding room IDs since they always contain a `:`, and we should always be query-encoding event IDs as they always contain `:` in room versions 1 and 2, and can potentially contain `=` and `/` in room version 3.

neilalexander:
Interestingly there's nothing in the spec that I can see that mentions it, either in the CS API, SS API or the appendices. I'll update this PR though so that we also query-encode room IDs throughout for compliance.

PR #186, opened 2020-04-30
Fix check on ExpiredTS

neilalexander:
If `ExpiredTS` is set then we should really return true if the supplied `atTS` is less than the expired time.


Kegsay:
Please add a test.

neilalexander:
@Kegsay
> Please add a test.

Done.

PR #189, opened 2020-05-04
Add more helper functions to Event

Kegsay:
Also fixes a bug in `Membership()` which would incorrectly return a value if there was no state key.

### DIFF
--- a/event.go
+++ b/event.go
@@ -841,11 +841,7 @@ func (e *Event) PrevEventIDs() []string {
 	}
 }
 
-// Membership returns the value of the content.membership field if this event
-// is an "m.room.member" event.
-// Returns an error if the event is not a m.room.member event or if the content
-// is not valid m.room.member content.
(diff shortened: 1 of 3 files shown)


event.go
```diff
@@ -859,16 +855,72 @@ func (e *Event) Membership() (string, error) {
 	default:
 		panic(e.invalidFieldType())
 	}
-	if fields.Type != MRoomMember {
-		return "", fmt.Errorf("gomatrixserverlib: not an m.room.member event")
+	if fields.Type != eventType {
+		return fmt.Errorf("gomatrixserverlib: not a %s event", eventType)
 	}
+	return json.Unmarshal(fields.Content, &content)
```
neilalexander, line 861: Should we check that `len(fields.Content) > 0`?

Kegsay: No, as it will just give an unmarshal error in that case.

PR #196, opened 2020-05-14
Implement key caching

neilalexander:
This moves in-memory key caching logic into gomatrixserverlib itself, so that it is consistent regardless of the backend database or fetchers used.


neilalexander:
So I did wonder about that, and I also created matrix-org/dendrite#1038 as an alternative approach. Pros for doing it in gmsl are that someone else can implement caching very easily, pros for doing it in Dendrite means no extra gmsl API surface. I'll happily go with whichever.

neilalexander:
Closing in favour of matrix-org/dendrite#1038.

PR #197, opened 2020-05-21
Allow marshalling and unmarshalling PublicKeyLookupRequest to/from text

neilalexander:
For the server key API component, it will make life a lot easier if we can continue to use `PublicKeyLookupRequest` as a map key, but we need to implement `encoding.TextMarshaler` and `encoding.TextUnmarshaler` so that `encoding/json` can encode and decode the map keys as valid JSON.

### DIFF
--- a/keyring.go
+++ b/keyring.go
@@ -2,6 +2,7 @@ package gomatrixserverlib
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


neilalexander:
According to RFC 952:
```
      <hname> ::= <name>*["."<name>]
      <name>  ::= <let>[*[<let-or-digit-or-hyphen>]<let-or-digit>]
```
In Matrix we don't currently do anything additional to that, so it should be fine. If it becomes a problem we can change the delimiter - the only place it's going to be used is in the internal HTTP APIs.

PR #198, opened 2020-05-28
Prevent NewEventFromUntrustedJSON from accepting headered events

neilalexander:
This should hopefully help us to avoid bugs like matrix-org/dendrite#1070.


event.go
```diff
@@ -262,6 +262,11 @@ func (eb *EventBuilder) Build(
 // It also checks the content hashes to ensure the event has not been tampered with.
 // This should be used when receiving new events from remote servers.
 func NewEventFromUntrustedJSON(eventJSON []byte, roomVersion RoomVersion) (result Event, err error) {
+	if r := gjson.GetBytes(eventJSON, "_*"); r.Exists() {
+		err = UnexpectedHeaderedEvent{}
```
Kegsay, line 266: Can we give context here (mention `NewEventFromUntrustedJSON`?)

neilalexander: You mean in the actual error text?

PR #200, opened 2020-06-02
Responsibility for marking room versions supported/stable

neilalexander:
Since gomatrixserverlib ultimately implements all of the logic related to specific room versions, it makes sense that it should also own the information about which versions are supported/stable.

### DIFF
--- a/eventversion.go
+++ b/eventversion.go
@@ -44,32 +44,42 @@ const (
 	StateResV2                              // state resolution v2
 )
 
-var roomVersionMeta = map[RoomVersion]roomVersion{
-	"1": roomVersion{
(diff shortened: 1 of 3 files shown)


eventversion.go
```diff
@@ -44,42 +44,94 @@ const (
...
+	}
+	return versions
+}
+
+// RoomVersionDescription contains information about a given room version, e.g. which
 // state resolution algorithm or event ID format to use.
-type roomVersion struct {
+// RoomVersionDescription contains information about a room version,
+// namely whether it is marked as supported or stable in this server
+// version, along with the state resolution algorith, event ID etc
```
Kegsay, line 124: > algorith

PR #201, opened 2020-06-03
Room version 6 support

neilalexander:
This implements the functionality required for room version 6 support. I'm not particularly thrilled about the reflection to verify the JSON int bounds but at this stage I'm not sure how better to handle it.

### DIFF
--- a/eventauth.go
+++ b/eventauth.go
@@ -465,6 +465,13 @@ func powerLevelsEventAllowed(event Event, authEvents AuthEventProvider) error {
 		return err
 	}
 
+	// Check that the changes in notification levels are allowed.
+	if notifs, err := event.roomVersion.PowerLevelsIncludeNotifications(); err == nil && notifs {
+		if err = checkNotificationLevels(senderLevel, oldPowerLevels, newPowerLevels); err != nil {
+			return err
+		}
+	}
+
 	// Check that the changes in user levels are allowed.
 	return checkUserLevels(senderLevel, event.Sender(), oldPowerLevels, newPowerLevels)
 }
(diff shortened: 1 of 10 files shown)


eventauth.go
```diff
@@ -630,6 +637,74 @@ func checkUserLevels(senderLevel int64, senderID string, oldPowerLevels, newPowe
...
+func checkNotificationLevels(senderLevel int64, oldPowerLevels, newPowerLevels PowerLevelContent) error {
+	type levelPair struct {
+		old    int64
+		new    int64
+		userID string
+	}
+
+	// First add the notification default level.
+	notificationLevelChecks := []levelPair{
+		//	{oldPowerLevels.UsersDefault, newPowerLevels.UsersDefault, ""},
```
Kegsay, line 650: Confused because we aren't adding a default

json_test.go
```diff
@@ -19,6 +19,31 @@ import (
 	"testing"
 )
 
+func TestJSONIntegerRanges(t *testing.T) {
+	// This value is in range so it should be fine with both room versions
+	input := `{"foo": 9007199254740991}`
+	if _, err := EnforcedCanonicalJSON([]byte(input), RoomVersionV1); err != nil {
+		// Room version 1 allows this value
+		t.Fatalf("should be valid")
```
Kegsay, line 27: Prefer `Errorf` here so it runs the other checks. Also please include the `err` in the log output.

json.go
```diff
@@ -27,14 +30,67 @@ import (
...
-		return nil, fmt.Errorf("invalid json")
+		return nil, BadJSONError{errors.New("gjson validation failed")}
 	}
 
 	return CanonicalJSONAssumeValid(input), nil
 }
 
+// Returns a gomatrixserverlib.BadJSONError if the canonical JSON fails enforced
+// checks or if JSON validation fails.
+func EnforcedCanonicalJSON(input []byte, roomVersion RoomVersion) ([]byte, error) {
```
Kegsay, line 44: Please link to what these checks are.

json.go
```diff
@@ -27,14 +30,67 @@ import (
...
+				if err := walk(v.MapIndex(k)); err != nil {
+					return err
+				}
+			}
+		case reflect.Int, reflect.Int64:
+			if v.Int() < -9007199254740991 || v.Int() > 9007199254740991 {
+				return fmt.Errorf("%d is outside of safe range", v.Int())
+			}
+		case reflect.Float64:
+			if v.Float() < -9007199254740991 || v.Float() > 9007199254740991 {
```
Kegsay, line 82: https://matrix.org/docs/spec/appendices#canonical-json indicates that floats are illegal.

PR #204, opened 2020-06-04
Re-request keys that have passed their validity lifetime

neilalexander:
When we retrieve a key from the database (or cache), then check the validity of the key. If we've passed the validity but the key is not yet expired, re-request it from the server to get the updated validity (or replacement keys).

### DIFF
--- a/keyring.go
+++ b/keyring.go
@@ -195,9 +195,10 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest) 
 
 	// TODO: we should distinguish here between expired keys, and those we don't have.
 	// If the key has expired, it's no use re-requesting it.
+	now := AsTimestamp(time.Now())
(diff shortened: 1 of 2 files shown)


keyring.go
```diff
@@ -195,8 +195,14 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest)
 
 	// TODO: we should distinguish here between expired keys, and those we don't have.
 	// If the key has expired, it's no use re-requesting it.
+	now := AsTimestamp(time.Now())
 	keysFetched := map[PublicKeyLookupRequest]PublicKeyLookupResult{}
 	for req, res := range keysFromDatabase {
+		if now > res.ValidUntilTS && res.ExpiredTS == PublicKeyNotExpired {
+			// We have passed the key validity period so we should re-request from
+			// the remote server (or the perspectives).
```
Kegsay, line 203: Please add a new test for this.

PR #205, opened 2020-06-05
Use AuthChainProvider for Check etc

neilalexander:
This uses `AuthChainProvider` as an optional parameter to:
- `RespState.Check()`
- `RespSendJoin.Check()`
- `checkAllowedByAuthEvents()`

The idea here is that the above functions don't just give up entirely if they hit a missing auth event, but instead can delegate to the provided function to *attempt* to retrieve the missing event. This means that we don't have to return to the caller and repeat the entire set of checks for each missing auth event.

If an `AuthChainProvider` isn't supplied then a `MissingAuthEventError` will be returned as usual.

### DIFF
--- a/authchain.go
+++ b/authchain.go
@@ -54,7 +54,7 @@ func VerifyEventAuthChain(ctx context.Context, eventToVerify HeaderedEvent, prov
 			eventsToVerify = append(eventsToVerify, newEvents...) // verify these events too
 		}
 		// verify the event
-		if err := checkAllowedByAuthEvents(curr, eventsByID); err != nil {
+		if err := checkAllowedByAuthEvents(curr, eventsByID, nil); err != nil {
 			return fmt.Errorf("gomatrixserverlib: VerifyEventAuthChain %v failed auth check: %w", curr.EventID(), err)
 		}
 		// add to the verified list
--- a/authchain.go
+++ b/authchain.go
@@ -54,7 +54,7 @@ func VerifyEventAuthChain(ctx context.Context, eventToVerify HeaderedEvent, prov
 			eventsToVerify = append(eventsToVerify, newEvents...) // verify these events too
 		}
 		// verify the event
-		if err := checkAllowedByAuthEvents(curr, eventsByID, nil); err != nil {
+		if err := checkAllowedByAuthEvents(curr, eventsByID, provideEvents); err != nil {
 			return fmt.Errorf("gomatrixserverlib: VerifyEventAuthChain %v failed auth check: %w", curr.EventID(), err)
 		}
 		// add to the verified list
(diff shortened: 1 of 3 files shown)


federationtypes.go
```diff
@@ -370,8 +370,14 @@ func (r RespState) Events() ([]Event, error) {
 	return result, nil
 }
 
+// MissingAuthEventHandler is a function signature which can be provided to
+// RespState.Check(), RespSendJoin.Check() and checkAllowedByAuthEvents. If
+// an auth event comes up missing during the checks, the MissingAuthEventHandler
+// will be called (if provided) to try and retrieve the missing event.
+type MissingAuthEventHandler func(eventID string) (*Event, error)
```
Kegsay, line 377: You'll probably need the room version as per https://github.com/matrix-org/gomatrixserverlib/blob/master/backfill.go#L26

neilalexander: Have added it for good measure, although the caller probably already knows it.

federationtypes.go
```diff
@@ -370,8 +370,14 @@ func (r RespState) Events() ([]Event, error) {
 	return result, nil
 }
 
+// MissingAuthEventHandler is a function signature which can be provided to
+// RespState.Check(), RespSendJoin.Check() and checkAllowedByAuthEvents. If
+// an auth event comes up missing during the checks, the MissingAuthEventHandler
+// will be called (if provided) to try and retrieve the missing event.
+type MissingAuthEventHandler func(eventID string, roomVersion RoomVersion) (*Event, error)
```
Kegsay, line 377: Tempted to just use https://github.com/matrix-org/gomatrixserverlib/blob/master/authchain.go#L9

authchain.go
```diff
@@ -54,7 +54,7 @@ func VerifyEventAuthChain(ctx context.Context, eventToVerify HeaderedEvent, prov
 			eventsToVerify = append(eventsToVerify, newEvents...) // verify these events too
 		}
 		// verify the event
-		if err := checkAllowedByAuthEvents(curr, eventsByID); err != nil {
+		if err := checkAllowedByAuthEvents(curr, eventsByID, nil); err != nil {
```
Kegsay, line 57: You can use `provideEvents` to implement the callback.

federationtypes.go
neilalexander: That's not a bad idea actually. Let me take a look.

federationtypes.go
```diff
@@ -590,19 +596,48 @@ type RespProfile struct {
...
+				// If we have a MissingAuthEventHandler then ask it for the missing event.
+				if ev, err := missingAuth(ae, event.roomVersion); err == nil {
+					// It claims to have returned an event - populate the eventsByID
+					// map so that we can retry with the new event.
+					eventsByID[ae] = ev
+				} else {
+					// It claims to have not returned an event - put a nil into the
+					// eventsByID map instead. This signals that we tried to retrieve
+					// the event but failed, so we don't keep retrying.
+					eventsByID[ae] = nil
```
Kegsay, line 617: This is a weird API shape. We don't use the `error` so just make the signature `missingAuth(event, ver) *Event` and always dump it into the map, they can return `nil` if they can't find it.

authchain.go
Kegsay: Please use authEventProvider here - `provideEvents`.

PR #206, opened 2020-06-08
Adjust DirectKeyFetcher

neilalexander:
This will:

- increase logging a bit when things go wrong fetching keys
- not re-request keys that are expired
- still allow updates that are past the validity to be returned, as `checkUsingKeys` will yell if there's a problem with them

### DIFF
--- a/keyring.go
+++ b/keyring.go
@@ -195,12 +195,12 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest) 
 
 	// TODO: we should distinguish here between expired keys, and those we don't have.
 	// If the key has expired, it's no use re-requesting it.
-	//now := AsTimestamp(time.Now())
 	keysFetched := map[PublicKeyLookupRequest]PublicKeyLookupResult{}
 	for req, res := range keysFromDatabase {
-		//if (now <= res.ValidUntilTS && res.ExpiredTS == PublicKeyNotExpired) || res.ExpiredTS != PublicKeyNotExpired {
-		//	delete(keyRequests, req)
-		//}
+		if res.ExpiredTS != PublicKeyNotExpired {
+			// Don't bother re-requesting keys that are marked as expired.
+			delete(keyRequests, req)
+		}
 		keysFetched[req] = res
 	}
 
--- a/keyring.go
+++ b/keyring.go
@@ -156,6 +156,11 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest) 
 	results := make([]VerifyJSONResult, len(requests))
 	keyIDs := make([][]KeyID, len(requests))
 
+	// Store the initial number of requests that were made. We'll remove
+	// things from the requests array that we no longer need, but we later
+	// need to check that we satisfied the full number of requests.
+	numRequests := len(requests)
+
 	for i := range requests {
 		ids, err := ListKeyIDs(string(requests[i].ServerName), requests[i].Message)
 		if err != nil {
@@ -193,9 +198,6 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest) 
 		return nil, err
 	}
 
-	// Default to not hitting federation.
-	doFederationHit := false
-
 	keysFetched := map[PublicKeyLookupRequest]PublicKeyLookupResult{}
 	now := AsTimestamp(time.Now())
 	for req, res := range keysFromDatabase {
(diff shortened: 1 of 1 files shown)


keyring.go
```diff
@@ -193,23 +193,29 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest)
...
+			// Don't bother re-requesting keys that are marked as expired.
+			delete(keyRequests, req)
+		}
 		if now > res.ValidUntilTS && res.ExpiredTS == PublicKeyNotExpired {
-			// We have passed the key validity period so we should re-request from
-			// the remote server (or the perspectives).
-			continue
+			// We're past the validity for this key so we should really
+			// hit the fetchers regardless to renew it.
+			doFederationHit = true
```
Kegsay, line 211: If this is `true` then why bother doing `checkUsingKeys` on `:216`? It's just wasted effort as it will never then return early.

keyring.go
```diff
@@ -231,8 +243,12 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest)
 
 		fetched, err := fetcher.FetchKeys(ctx, keyRequests)
 		if err != nil {
-			fetcherLogger.WithError(err).WithField("fetcher", fetcher.FetcherName()).
-				Warn("Failed to request keys from fetcher")
+			fetcherLogger.WithError(err).Warn("Failed to request keys from fetcher")
```
Kegsay, line 246: We've lost the fetcher name here.

neilalexander: `fetcherLogger` already includes that field, it was duplicated in this case.

keyring.go
```diff
@@ -193,23 +193,29 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest)
...
+	doFederationHit := false
+
+	// TODO: we should distinguish here between expired keys, and those we
+	// don't have. If the key has expired, it's no use re-requesting it.
 	keysFetched := map[PublicKeyLookupRequest]PublicKeyLookupResult{}
+	now := AsTimestamp(time.Now())
 	for req, res := range keysFromDatabase {
+		if res.ExpiredTS != PublicKeyNotExpired {
+			// Don't bother re-requesting keys that are marked as expired.
+			delete(keyRequests, req)
```
Kegsay, line 206: This is implementing the TODO isn't it?
```
// TODO: we should distinguish here between expired keys, and those we
// don't have. If the key has expired, it's no use re-requesting it.
```
so remove the TODO comment?

keyring.go
Kegsay: I don't see how setting this to true helps - if the key is bad the check will fail and THEN we'll hit federation?

neilalexander: I've shuffled this around a bit now - we'll only try the first `checkUsingKeys` if we think all of our key requests are satisfied. Then if we *still* need to hit federation, we will do.

Kegsay: I don't see what `doFederationHit` buys us in this part of the code. It would be just as good to say `doFederationHit := len(keysFetched) < len(keyRequests)` which is also clearer IMO.

keyring.go
```diff
@@ -193,34 +193,57 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest)
...
+	if len(keysFetched) == len(keyRequests) {
+		// If our key requests are all satisfied then we can try performing
+		// a verification using our keys.
+		k.checkUsingKeys(requests, results, keyIDs, keysFetched)
+
+		// If we run into any errors when verifying using the keys that we
+		// have then we can hit federation and check for updated keys.
+		for _, r := range results {
+			if r.Error != nil {
+				doFederationHit = true
```
Kegsay, line 229: I would combine this with `if !doFederationHit { ... }` so in other words what you're saying is:
- If `len(keysFetched) == len(keyRequests)` and there's no errors from `checkUsingKeys` then return results
- Else do federation hits.

Basically we don't need `doFederationHit`.

neilalexander: The logic here is:

1. Ask for keys that we know about from the database:
    1. If *any of the keys* from the database are past their validity then we need to hit federation anyway in order to get the renewed validities.
2. If we successfully satisfy all requested keys, try to `checkWithKeys`:
    1. If it succeeds (i.e. we haven't claimed we need to hit federation), return the results and stop here.
    2. If it doesn't succeed, hit federation in case one of the keys isn't what we thought it was.
3. If we didn't successfully satisfy all requested keys, then hit federation to get the keys we don't know about yet.

I'm not sure removing `doFederationHit` makes that pattern any clearer.

Kegsay: I think it does maker it clearer overall. Then you can just do:
```go
if len(keysFetched) == len(keyRequests) {
    k.checkUsingKeys(requests, results, keyIDs, keysFetched)
    noErrors := true
    for _, r := range results {
        if r.Error != nil {
            noErrors := false
            break
        }
    }
    if noErrors {
        return results, nil
    }
}

// hit federation
```

neilalexander: Have done that as a part of some greater magnifying-glass-ing, and I've also made one or two more tweaks. As far as I can tell from copious logging, this *should* be doing the right thing now.

PR #214, opened 2020-08-07
Allow enforcing X.509 certificate validity (MSC1711)

neilalexander:
This makes skipping certificate validation optional so that we can sort matrix-org/dendrite#760.

### DIFF
--- a/client.go
+++ b/client.go
@@ -48,12 +48,12 @@ type UserInfo struct {
 }
 
 // NewClient makes a new Client (with default timeout)
-func NewClient() *Client {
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -96,7 +98,7 @@ func (f *federationTripper) getTransport(tlsServerName string) (transport http.R
 			TLSClientConfig: &tls.Config{
 				ServerName: tlsServerName,
 				// TODO: Remove this when we enforce MSC1711.
```
Kegsay, line 100: Comment is obsolete now

PR #219, opened 2020-09-03
add peek constant

ara4n:
needed for peeking (even though it's not technically a membership state, we use it in dendrite as the key for the appropriate sync section for [MSC2753](https://github.com/matrix-org/matrix-doc/pull/2753) style peeking)

### DIFF
--- a/eventauth.go
+++ b/eventauth.go
@@ -34,7 +34,7 @@ const (
 	Leave = "leave"
 	// Invite is the string constant "invite"
 	Invite = "invite"
-	// Peek is the string constant "peek"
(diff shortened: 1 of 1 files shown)


eventauth.go
```diff
@@ -34,6 +34,8 @@ const (
 	Leave = "leave"
 	// Invite is the string constant "invite"
 	Invite = "invite"
+	// Peek is the string constant "peek"
```
neilalexander, line 37: Would be good if you can `// NOTSPEC:` this comment and add a link to the MSC.

neilalexander (approved):

PR #220, opened 2020-09-04
WIP: MSC2444 /peek support in the federation client

ara4n:


federationtypes.go
```diff
@@ -204,6 +219,16 @@ type respStateFields struct {
 	AuthEvents  []Event `json:"auth_chain"`
 }
 
+// XXX: this duplicates RespPeek above and needs to be kept in sync with it
```
neilalexander, line 222: It looks like you’ve tried to copy a bit of a Go hack — it lets you define a bunch of public fields (so that you can JSON marshal/unmarshal), but because they are in an unexported struct, it prevents things from outside the package manipulating those fields directly.

Normally if you do this, `RespPeek` would just embed the private struct (`type RespPeek struct { respPeekFields }`) but might require some marshalling/unmarshalling wrappers, otherwise you can probably just get rid of `respPeekFields` if it isn’t being used anywhere else deliberately.

federationtypes.go
```diff
@@ -256,6 +281,60 @@ func (r *RespMissingEvents) UnmarshalJSON(data []byte) error {
...
+		r.AuthEvents = []Event{}
+	}
+	return json.Marshal(respPeekFields(r))
+}
+
+// UnmarshalJSON implements json.Unmarshaller
+func (r *RespPeek) UnmarshalJSON(data []byte) error {
+	r.AuthEvents = []Event{}
+	r.StateEvents = []Event{}
+	// XXX: it feels broken that we have RespPeek, respPeekFields and intermediate
```
neilalexander, line 299: The key difference in the `intermediate` stage is that the state events and auth events are not unmarshalled on the first pass (hence `json.RawMessage`), which allows you to get `RoomVersion` so that you know what event format to unmarshal them as.

client.go
```diff
@@ -333,6 +333,10 @@ func (fc *Client) DoRequestAndParseResponse(
 		return err
 	}
 
+	// body, err := ioutil.ReadAll(response.Body)
```
neilalexander, line 336: Would be good to remove the dead code if it isn’t needed anymore.

federationtypes.go
neilalexander: (In this instance, having internal fields might be overkill so probably best just to remove that, otherwise you also need to define exported functions to get things out of the struct.)

PR #221, opened 2020-09-07
Don't cause panics when trimming failed auth/state events

neilalexander:
This happened because we were trimming things out of the slices but the original `range` was running on a copy of the slice, meaning that we'd hit an invalid index and panic.

This helps matrix-org/dendrite#1401 — sort of. Joining that room still doesn't work because of a bad state event in the room, but at least we don't panic anymore.


Kegsay (approved):
Subtle!

PR #223, opened 2020-09-23
Omit tls_fingerprints from server keys when empty

tulir:
Construct breaks if it's null instead of an array. It's a legacy field which shouldn't be used anyway, so it could be removed altogether, but this is just a quick fix that shouldn't affect anything else.

Signed-off-by: Tulir Asokan &lt;tulir@maunium.net&gt;


neilalexander:
Thanks for this!

PR #226, opened 2020-10-01
Remove event ID-origin check, move to signature checking

neilalexander:
This removes the check for the event ID domain and the origin field being the same, instead moving the responsibility onto checking that the event ID domain has provided a signature (on room versions where it is applicable).

This means that we now do the same as Synapse.

### DIFF
--- a/eventcrypto.go
+++ b/eventcrypto.go
@@ -243,6 +243,19 @@ func VerifyEventSignatures(ctx context.Context, events []Event, keyRing JSONVeri
 		}
 		senderServer := ServerName(senderDomain)
 		domains[senderServer] = true
+		// Synapse requires that the event ID domain has a valid signature.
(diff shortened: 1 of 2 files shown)


eventcrypto.go
```diff
@@ -243,6 +243,19 @@ func VerifyEventSignatures(ctx context.Context, events []Event, keyRing JSONVeri
 		}
 		senderServer := ServerName(senderDomain)
 		domains[senderServer] = true
+		// Synapse requires that the event ID domain has a valid signature.
+		// https://github.com/matrix-org/synapse/blob/develop/synapse/event_auth.py#L98-L104
+		// There's a good reason for this. Surely.
```
Kegsay, line 248: I think so you can't spoof event ID domains on event IDs?

Kegsay (approved):

eventcrypto.go
neilalexander: Yes, I imagine so.

PR #231, opened 2020-10-05
Parallelise checkUsingKeys

neilalexander:
This speeds up signature checks by performing them across available CPU cores (less one, to not totally starve the system).

This might cause problems if lots of big signature checks are taking place concurrently though, so there's that.

### DIFF
--- a/keyring.go
+++ b/keyring.go
@@ -347,7 +347,7 @@ func (k *KeyRing) checkUsingKeys(
 	wg.Add(len(jobs))
 	for _, j := range jobs {
 		go func(jobs []job) {
-			for i, j := range jobs {
+			for _, j := range jobs {
 				mu.RLock()
 				if results[j.index].Error == nil {
 					// We've already checked this message and it passed the signature checks.
--- a/keyring.go
+++ b/keyring.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"runtime"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


keyring.go
```diff
@@ -325,39 +326,69 @@ func (k *KeyRing) checkUsingKeys(
...
+					}
+					if !serverKey.WasValidAt(j.request.AtTS, j.request.StrictValidityChecking) {
+						// The key wasn't valid at the timestamp we needed it to be valid at.
+						// So skip onto the next key.
+						mu.Lock()
+						results[j.index].Error = fmt.Errorf(
+							"gomatrixserverlib: key with ID %q for %q not valid at %d",
+							keyID, j.request.ServerName, j.request.AtTS,
+						)
+						mu.Unlock()
```
Kegsay, line 369: I'm not a fan of the shared variables across goroutines, it feels like a code smell as opposed to sharing data via channels. I'm aware we do this to maintain the index in the results array.

I think a better solution would be to add the parallelisation around `VerifyJSON` only, meaning you just need to pass information that is directly used without needing to access shared slices/maps. Something like:

```go
type verifyJSONReq struct {
    i int
    serverName ServerName
    keyID KeyID
    pubKey ed25519.PublicKey
    msg []byte
}
type verifyJSONRes struct {
    i int
    err error
}
reqCh := make(chan verifyJSONReq)
resCh := make(chan verifyJSONRes)
var wg sync.WaitGroup
wg.Add(procs)
for i := 0; i < procs; i++ {
    go func() {
        defer wg.Done()
        for item := range reqCh {
             err := VerifyJSON(item.serverName, item.keyID, item.pubKey, item.msg)
            resCh <- verifyJSONRes{item.i, err}
        }
    }
}
for i := range requests {
    // insert code that was there previously up to VerifyJSON
    reqCh <- verifyJSONReq{i, other, func, args}
}
close(reqCh) // kill the goroutines after they process everything
// kill the response channel when we've got all the response
go func() {
    wg.Wait()
    close(resCh)
}()
for res := range resCh {
    results[res.i].Error = res.err
}
```

PR #237, opened 2020-10-15
State res v2 optimisations

neilalexander:
This makes a number of optimisations in the state res v2 code, largely around preallocating backing arrays, identifying unconflicted events using a map rather than loops, hitting some common event functions less often.

### DIFF
--- a/stateresolutionv2.go
+++ b/stateresolutionv2.go
@@ -543,7 +543,7 @@ func (r *stateResolverV2) getPowerLevelFromAuthEvents(event *Event) (pl int) {
 func kahnsAlgorithmUsingAuthEvents(events []*stateResV2ConflictedPowerLevel) (
 	graph []*stateResV2ConflictedPowerLevel,
 ) {
-	eventMap := make(map[string]*stateResV2ConflictedPowerLevel)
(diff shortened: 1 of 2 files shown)


stateresolutionv2.go
```diff
@@ -374,49 +374,51 @@ func (r *stateResolverV2) authAndApplyEvents(events []*Event) {
...
 		case MRoomCreate:
 			// Room creation events are only valid with an empty state key.
-			if event.StateKey() == nil || *event.StateKey() == "" {
+			if sk == nil || *sk == "" {
 				r.resolvedCreate = event
 			}
 		case MRoomPowerLevels:
 			// Power level events are only valid with an empty state key.
-			if event.StateKey() == nil || *event.StateKey() == "" {
+			if sk == nil || *sk == "" {
```
Kegsay, line 387: This surely makes it possible for an `m.room.power_levels` event with no state key (aka a message event) pass these checks? We should be checking `if sk != nil && *sk == ""`

Kegsay (approved):

stateresolutionv2.go
neilalexander: Whoops, good catch. Thanks.

PR #242, opened 2020-11-20
MSC2836 request

Kegsay:


federationtypes.go
```diff
@@ -873,3 +874,51 @@ type DeviceKeys struct {
...
+	r.RecentFirst = true
+	r.IncludeParent = false
+	r.IncludeChildren = false
+	r.Direction = "down"
+}
+
+// MSC2836EventRelationshipsResponse is a response to /event_relationships from
+// https://github.com/matrix-org/matrix-doc/blob/kegan/msc/threading/proposals/2836-threading.md
+type MSC2836EventRelationshipsResponse struct {
+	Events    []Event `json:"events"`
```
neilalexander, line 920: When a response like this comes in from the wire, we need to know the room version ahead of time before unmarshalling `[]Event`.

That's why `RespMissingEvents`, `RespPeek`, `RespState`, `RespSendJoin` etc have the intermediate unmarshalling step where you take out the events as `json.RawMessage` and then run them through `NewEventFromUntrustedJSON` with the room version that was prepopulated from somewhere.

I suspect you'll need to do something similar here, otherwise the unmarshaller won't know which event format to pick.

neilalexander (approved):
Looks good to me!

PR #243, opened 2020-12-03
Add NewEventFromTrustedJSONWithEventID

neilalexander:
This adds `NewEventFromTrustedJSONWithEventID`, which is used specifically when we're pulling the event out of storage and *we already know the event ID*. This saves us *a lot* of CPU time because then we don't need to calculate the reference hash, redact, canonicalise etc just to work it out.


Kegsay (approved):
Looks good!

PR #250, opened 2021-01-21
Limit well-known response sizes

neilalexander:
This prevents us from pulling in potentially huge well-known files and using lots of memory.

cc @clokep for info.

### DIFF
--- a/well_known.go
+++ b/well_known.go
@@ -3,7 +3,8 @@ package gomatrixserverlib
 import (
 	"encoding/json"
 	"errors"
-	"io/ioutil"
+	"fmt"
+	"io"
 	"net/http"
 	"strconv"
 	"strings"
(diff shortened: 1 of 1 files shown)


well_known.go
```diff
@@ -79,10 +89,20 @@ func LookupWellKnown(serverNameType ServerName) (*WellKnownResult, error) {
...
 	}
 
-	body, err := ioutil.ReadAll(resp.Body)
+	// By this point we hope that we've caught any huge well-known records
+	// by checking Content-Length, but it's possible that header will be
+	// missing. Better to be safe than sorry by reading no more than the
+	// WellKnownMaxSize in any case.
+	bodyBuffer := make([]byte, WellKnownMaxSize)
+	limitedReader := &io.LimitedReader{
+		R: resp.Body,
```
Kegsay, line 98: We still need to close the body at some point.

Kegsay (approved):

well_known.go
neilalexander: There's a `defer` further up the function:
```
	defer func() {
		_ = resp.Body.Close()
	}()
```

PR #251, opened 2021-01-21
DNS caching for federation traffic

neilalexander:
This stops Dendrite from hitting DNS so much.

### DIFF
--- a/client.go
+++ b/client.go
@@ -113,7 +113,7 @@ func (f *federationTripper) getTransport(tlsServerName string) (transport http.R
 		if f.dnsCache != nil {
 			tr.DialContext = f.dnsCache.DialContext
 		}
-		f.transports[tlsServerName] = tr
+		transport, f.transports[tlsServerName] = tr, tr
 	}
 
 	f.transportsMutex.Unlock()
--- a/client.go
+++ b/client.go
@@ -79,12 +79,14 @@ type federationTripper struct {
 	transportsMutex sync.Mutex
 	skipVerify      bool
 	resolutionCache sync.Map // serverName -> []ResolutionResult
+	dnsCache        *dnsCache
 }
 
 func newFederationTripper(skipVerify bool) *federationTripper {
 	return &federationTripper{
 		transports: make(map[string]http.RoundTripper),
 		skipVerify: skipVerify,
+		dnsCache:   newDNSCache(1024, time.Minute),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -79,12 +79,21 @@ type federationTripper struct {
...
+	dnsCache        *DNSCache
 }
 
-func newFederationTripper(skipVerify bool) *federationTripper {
+func newFederationTripper(skipVerify bool, options ...interface{}) *federationTripper {
+	var dnsCache *DNSCache
+	for _, opt := range options {
+		switch o := opt.(type) {
+		case *DNSCache:
+			dnsCache = o
```
Kegsay, line 90: Generally the idiom is to use `WithDNSCache(thing)` instead of this. Can we make it use that shape instead?

dnscache.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+type dnsCacheEntry struct {
+	addrs   []net.IPAddr
+	expires time.Time
+}
+
+func (c *DNSCache) lookup(ctx context.Context, name string) (*dnsCacheEntry, bool) {
+	c.mutex.Lock()
+	defer c.mutex.Unlock()
```
Kegsay, line 35: Should we be holding onto the lock during `LookupIPAddr`?

dnscache.go
```diff
@@ -0,0 +1,130 @@
...
+	resolver *net.Resolver
+	mutex    sync.Mutex
+	size     int
+	duration time.Duration
+	entries  map[string]*dnsCacheEntry
+}
+
+func NewDNSCache(size int, duration time.Duration) *DNSCache {
+	return &DNSCache{
+		resolver: net.DefaultResolver,
```
Kegsay, line 21: Please can you add some unit tests for the `DNSCache` e.g replacing `resolver *net.Resolver` with `resolver netResolver` where:
```go
type netResolver interface {
    LookupIPAddr(ctx, name)
}
```
which allows us to stub out the network operations. Similarly, you'll need to do this for `dialer := net.Dialer{}` and `DialContext`.

dnscache.go
neilalexander: Oh, that's a good catch, I'll fix that.

dnscache.go
neilalexander: Done - added tests.

client.go
neilalexander: It is done with a `WithDNSCache` struct.

Kegsay (approved):

PR #252, opened 2021-01-22
Simplify NewClient and NewFederationClient with ClientOpts

neilalexander:
This removes some of the duplicate messy functions and returns us to just having `NewClient` and `NewFederationClient`. You can provide one or more `ClientOpts` to control the transport, timeout, TLS validation, DNS cache etc.

### DIFF
--- a/client.go
+++ b/client.go
@@ -58,7 +58,7 @@ type clientOptions struct {
 // ClientOption are supplied to NewClient or NewFederationClient.
 type ClientOption func(*clientOptions)
 
-// NewClient makes a new Client. You can supply zero or more ClientOpts
+// NewClient makes a new Client. You can supply zero or more ClientOptions
 // which control the transport, timeout, TLS validation etc - see
 // WithTransport, WithTimeout, WithSkipVerify, WithDNSCache etc.
 func NewClient(options ...ClientOption) *Client {
--- a/client.go
+++ b/client.go
@@ -52,7 +52,9 @@ type ClientOption interface {
 	IsClientOption()
 }
 
-// NewClient makes a new Client (with default timeout)
+// NewClient makes a new Client. You can supply zero or more ClientOpts
+// which control the transport, timeout, TLS validation etc - see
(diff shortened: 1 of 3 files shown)


client.go
```diff
@@ -52,37 +52,66 @@ type ClientOption interface {
...
+// WithTransport, WithTimeout, WithSkipVerify, WithDNSCache etc.
+func NewClient(options ...ClientOption) *Client {
+	var transport http.RoundTripper
+	var dnsCache *DNSCache
+	timeout := requestTimeout
+	skipVerify := false
+	for _, option := range options {
+		switch o := option.(type) {
+		case WithTransport:
+			transport = o.Transport
```
Kegsay, line 66: Can we not do this please? Can we just use `WithTransport` the Go way as per https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis tl;dr

```go
type ClientOption func(*Client)

func NewClient(options ...ClientOption) *Client {
    cli := &Client{}
    for _, opt := range options {
        opt(cli)
    }
}

func WithTransport(tr http.Transport) ClientOption {
    return func(cli *Client) {
        cli.Transport = tr
    }
}
```

federationclient.go
```diff
@@ -23,40 +21,16 @@ type FederationClient struct {
 	serverPrivateKey ed25519.PrivateKey
 }
 
-// NewFederationClient makes a new FederationClient
+// NewFederationClient makes a new FderationClient. You can supply
```
Kegsay, line 24:
```suggestion
// NewFederationClient makes a new FederationClient. You can supply
```

Kegsay (approved):

client.go
neilalexander: Well, I can shuffle it around to use functions instead of structs, that's not difficult, but we'll still need to write to an options struct before building the `&Client`/`http.Client` because `http.RoundTripper` is an opaque type, and we need to not be sensitive to the ordering of options.

PR #256, opened 2021-03-05
Don't default to federation tripper for normal Client

neilalexander:
`FederationClient` should default to the federation tripper but a normal `Client` shouldn't.

### DIFF
--- a/client.go
+++ b/client.go
@@ -68,6 +68,14 @@ func NewClient(options ...ClientOption) *Client {
 	for _, option := range options {
 		option(clientOpts)
 	}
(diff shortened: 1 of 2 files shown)


Kegsay (approved):
I really don't get why we want to use GMSL.Client but sure, this diff looks fine.

neilalexander:
Trying to fix the AS transports this way breaks the P2P demos so I'm just going to close this and go back to having the appservice manage its own `http.Client`s, which is *also* rubbish but apparently unavoidable.

PR #259, opened 2021-05-25
Make optional params from MSC2946 actually optional

babolivier:
Add the `omitempty` annotation to `MSC2946SpacesRequest.ExcludeRooms` and `MSC2946SpacesRequest.MaxRoomsPerSpace` as MSC2946 states these are optional parameters (and keeping them around with their zero value might lead to type errors when interacting with other homeservers). See https://github.com/matrix-org/synapse/issues/10056 for more context.


neilalexander (approved):

babolivier:
I've also noticed that the `suggested_only` optional parameter isn't supported, though it felt like a bit of scope creep to add it here.

PR #260, opened 2021-06-10
Use "[]byte" instead of "json.RawMessage" if posible

6543:
if merged we can use it in https://github.com/matrix-org/dendrite/pull/1870

but it's **BREAKING** change :/ - it's easy to resolve but need code change ...

Signed-off-by: 6543 <6543@obermui.de>


6543:
to de-serialize  json slice inside a json field without using json.RawMessage ... is tricy ... I did not found a nice solution *yet* ... so this pull is more a proposal ...

kegsay:
Why is this desirable? The data is raw JSON which we don't yet know how to deserialise, so the data type is fine as it is.

kegsay:
`json.RawMessage` is just an alias for `[]byte` so it's not hard to typecast to it from another JSON library.

PR #268, opened 2021-07-14
state_default is 50 not 0

kegsay:
This is to address a few sytests, notably `Remote servers cannot set power levels in rooms without existing powerlevels` which checks that if a room has no PL event then a random user can't then set the PLs, it has to be the creator.

This change fixes the issue as if there is no PL event then the room creator gets a PL of 100, and everyone else is 0. Making the state_default 50 then means randoms cannot set the PL. See comments for relevant synapse code.


neilalexander (approved):
Huh, interesting.

ara4n:
can we tie this back to the spec rather than the synapse impl?

ara4n:
specifically in https://matrix.org/docs/spec/client_server/r0.6.1#m-room-power-levels:

Screenshot 2021-07-14 at 15 27 56

PR #269, opened 2021-07-19
Improve state resolution v2 performance using allowerContext

neilalexander:
This improves state resolution v2 performance quite dramatically in large rooms when authing and applying events by reducing the number of times that we parse the content of the create, power levels and join rule events from the auth event provider.

### DIFF
--- a/eventauth.go
+++ b/eventauth.go
@@ -336,10 +336,10 @@ func errorf(message string, args ...interface{}) error {
 	return &NotAllowed{Message: fmt.Sprintf(message, args...)}
 }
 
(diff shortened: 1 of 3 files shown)


kegsay (approved):
Is there any risk that the PL / join rule event may change between calls to `allowed`?

neilalexander:
In the state res v2 path, no, because by that point we've done the Kahn's and/or power level mainline ordering. In other cases it might if used incorrectly, but I've specifically not exported this codepath for that reason — it might create a footgun if it were used otherwise.

PR #270, opened 2021-07-20
Use intermediate step for RespSendJoin to drop invalid events

neilalexander:
This should fix matrix-org/dendrite#1927.

### DIFF
--- a/federationclient.go
+++ b/federationclient.go
@@ -110,7 +110,27 @@ func (ac *FederationClient) SendJoin(
 	if err = req.SetContent(event); err != nil {
 		return
 	}
-	err = ac.doRequest(ctx, req, &res)
+	var intermediate struct {
+		StateEvents []json.RawMessage `json:"state"`
+		AuthEvents  []json.RawMessage `json:"auth_chain"`
+		Origin      ServerName        `json:"origin"`
+	}
+	process := func() {
+		for _, se := range intermediate.StateEvents {
+			if ev, err := NewEventFromUntrustedJSON(se, roomVersion); err == nil {
+				res.StateEvents = append(res.StateEvents, ev)
+			}
+		}
+		for _, ae := range intermediate.AuthEvents {
+			if ev, err := NewEventFromUntrustedJSON(ae, roomVersion); err == nil {
+				res.AuthEvents = append(res.AuthEvents, ev)
+			}
+		}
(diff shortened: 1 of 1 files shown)


federationclient.go
```diff
@@ -110,7 +110,27 @@ func (ac *FederationClient) SendJoin(
...
-	err = ac.doRequest(ctx, req, &res)
+	var intermediate struct {
+		StateEvents []json.RawMessage `json:"state"`
+		AuthEvents  []json.RawMessage `json:"auth_chain"`
+		Origin      ServerName        `json:"origin"`
+	}
+	process := func() {
+		for _, se := range intermediate.StateEvents {
+			if ev, err := NewEventFromUntrustedJSON(se, roomVersion); err == nil {
+				res.StateEvents = append(res.StateEvents, ev)
```
kegsay, line 121: Given how large these arrays can be, maybe allocate a capacity to `len(intermediate.StateEvents)` to avoid too much churn?

federationclient.go
```diff
@@ -124,7 +144,10 @@ func (ac *FederationClient) SendJoin(
 		var v1Res []json.RawMessage
 		err = ac.doRequest(ctx, v1req, &v1Res)
 		if err == nil && len(v1Res) == 2 {
-			err = json.Unmarshal(v1Res[1], &res)
+			if err = json.Unmarshal(v1Res[1], &res); err != nil {
```
kegsay, line 147: Shouldn't this be unmarshalling to `intermediate`? It's kinda gross that `process()` takes args by side-effects as opposed to just passing them in which may have caught this? Can we modify `process` to accept `intermediate` as an arg rather than doing this in closures perhaps?

kegsay (approved):
Otherwise LGTM

federationclient.go
neilalexander: Good point, done.

federationclient.go
neilalexander: The main reason I didn't do this was because it avoided having to give a named type to `intermediate`, but yes, you're right that it is unmarshalling to the wrong place there for the fallback. I'll fix it.

kegsay: You don't _have_ to give named types, `func(a struct{Foo string}) { ... }` works. Kinda gross though.

PR #279, opened 2021-11-02
Room version 8 and 9 support

neilalexander:
In theory, this PR adds support for room version 8 and 9.


eventauth.go
```diff
@@ -976,6 +994,58 @@ func (m *membershipAllower) membershipAllowed(event *Event) error { // nolint: g
...
+	if err != nil {
+		return fmt.Errorf("failed to find the membership status for 'join_authorised_via_users_server' user %q", m.newMember.AuthorisedVia)
+	}
+	if otherMembership != Join {
+		return fmt.Errorf("the nominated 'join_authorised_via_users_server' user %q is not joined to the room", m.newMember.AuthorisedVia)
+	}
+
+	// And secondly, does the user have the power to issue invites in the room?
+	if m.powerLevels.UserLevel(m.newMember.AuthorisedVia) < m.powerLevels.Invite {
+		return fmt.Errorf("the nominated 'join_authorised_via_users_server' user %q does not have permission to invite", m.newMember.AuthorisedVia)
```
kegsay, line 1040: maybe include `%d < %d`.

eventauth.go
```diff
@@ -976,6 +994,58 @@ func (m *membershipAllower) membershipAllowed(event *Event) error { // nolint: g
...
+		m.joinRule.JoinRule = Invite
+		return nil
+	}
+
+	// Otherwise, we have to work out if the server that produced the join was
+	// authorised to do so. This requires the membership event to contain a
+	// 'join_authorised_via_users_server' key, containing the user ID of a user
+	// in the room that should have a suitable power level to issue invites.
+	// If no such key is specified then we should reject the join.
+	if _, _, err := SplitID('@', m.newMember.AuthorisedVia); err != nil {
```
kegsay, line 1020: What's stopping servers from lying about who authorised this? Can I not just be malicious and guess that "hey Alice is probably in $secret_room, let's slap an authorised key as alice and I can get in"?

kegsay (approved):

eventauth.go
neilalexander: Well, the authorising server has to have signed the event as well. I assume the malicious person could enter a different user ID from the same signing server, but unless that user has the power levels to invite, I don't suppose it'd help them at all.

PR #281, opened 2021-11-12
Refactor `SignJSON`, preserve signatures in `EventBuilder`

neilalexander:
This PR refactors `SignJSON` to just be so much less of a mess by using `gjson`/`sjson` instead, and also ensures that old signatures are preserved when we unmarshal an event into an `EventBuilder`.

### DIFF
--- a/event.go
+++ b/event.go
@@ -259,8 +259,10 @@ func (eb *EventBuilder) Build(
 		return
 	}
 
-	if eventJSON, err = sjson.SetBytes(eventJSON, "signatures", event.Signatures); err != nil {
-		return
+	if len(event.Signatures) > 0 {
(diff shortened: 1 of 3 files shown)


event.go
```diff
@@ -88,6 +88,8 @@ type EventBuilder struct {
 	// The depth of the event, This should be one greater than the maximum depth of the previous events.
 	// The create event has a depth of 1.
 	Depth int64 `json:"depth"`
+	// The JSON object for "signature" key of the event.
+	Signature RawJSON `json:"signatures,omitempty"`
```
kegsay, line 92: JSON tag is `signatures` but comment is `signature`?

kegsay (approved):

PR #288, opened 2022-02-08
Add EventJSON and EventJSONs types

kegsay:
This is largely driven by the desire to use GMSL Response types without knowing the room version up front. Now basically every response just returns these new types which then have functions to deserialise them into events with the given room version.


kegsay (commented):
+112 -335

We're doing something right ;)

neilalexander (approved):

neilalexander (approved):

PR #289, opened 2022-02-13
Use http.DefaultTransport, which loads proxy config from environment

S7evinK:
Should make the [proxy_outbound](https://github.com/matrix-org/dendrite/blob/5106cc807cf22a95420b24f6bfdd5c9ac8aa06de/dendrite-config.yaml#L207-L212) config obsolete, since it uses the default mechanism to get the proxy settings.


kegsay (approved):

neilalexander (approved):

neilalexander:
We should make a follow-up PR to remove the proxy section from the Dendrite config file.

PR #291, opened 2022-02-14
Add support for partial state in `send_join` responses

richvdh:
[MSC3706](https://github.com/matrix-org/matrix-doc/pull/3706) adds some new fields to `/send_join` responses. Here we add support for the new fields, and the new query param

### DIFF
--- a/federationtypes_test.go
+++ b/federationtypes_test.go
@@ -2,7 +2,11 @@ package gomatrixserverlib
 
 import (
 	"encoding/json"
+	"strings"
 	"testing"
+	"unicode"
+
+	"github.com/google/go-cmp/cmp"
 )
 
 const emptyRespStateResponse = `{"pdus":[],"auth_chain":[]}`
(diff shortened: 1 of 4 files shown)


federationtypes_test.go
```diff
@@ -90,38 +94,68 @@ func TestRespStateUnmarshalJSON(t *testing.T) {
...
 }
 
-func TestRespSendJoinUnmarshalJSON(t *testing.T) {
-	inputData := `{"state":[],"auth_chain":[],"origin":""}`
+func TestRespSendJoinMarshalJSONPartialState(t *testing.T) {
+	inputData := `{
+        "state":[],"auth_chain":[],"origin":"o1",
+	    "org.matrix.msc3706.partial_state":true,
+        "org.matrix.msc3706.servers_in_room":["s1", "s2"]
+    }`
```
squahtx, line 128: nit: strange indentation here. the middle line has a mix of spaces and tabs.

squahtx (approved):

federationtypes_test.go
richvdh: thanks. fixed.

PR #296, opened 2022-03-10
Implement `destination` field in `X-Matrix` header as per MSC3383

neilalexander:
This sets the `destination=` field in the federation request auth header so that a proxy can distinguish the target of a federation request without needing to rely on the `Host` header, which is more likely to be set to the hostname of the proxy itself. This should be useful for P2P.

### DIFF
--- a/request_test.go
+++ b/request_test.go
@@ -19,6 +19,7 @@ const exampleGetRequest = "GET /_matrix/federation/v1/query/directory?room_alias
 	" origin=\"localhost:8800\"" +
 	",key=\"ed25519:a_Obwu\"" +
 	",sig=\"7vt4vP/w8zYB3Zg77nuTPwie3TxEy2OHZQMsSa4nsXZzL4/qw+DguXbyMy3BF77XvSJmBt+Gw+fU6T4HId7fBg\"" +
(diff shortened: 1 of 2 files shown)


request_test.go
```diff
@@ -31,6 +32,7 @@ const examplePutRequest = "PUT /_matrix/federation/v1/send/1493385816575/ HTTP/1
 	" origin=\"localhost:8800\"" +
 	",key=\"ed25519:a_Obwu\"" +
 	",sig=\"+hmW6UjEXx7vMt2+MXO/EImSfdEYdBsZEOmpiz3evYktAgGNpGuNMBYXIA969WGubmceREKA/r1phasUFHBpDg\"" +
+	",destination=\"localhost:44033\"" +
```
kegsay, line 35: Are there some test cases with this field missing? Why add them here?

kegsay (approved):
We need to ensure that requests missing this field are okay too.

request_test.go
neilalexander: These tests do an equality check apparently, so the value is needed. If it's missing, we ignore it and populate it ourselves with the `serverName`, so it should be fine.

PR #297, opened 2022-03-11
`Client` and federation tripper improvements

neilalexander:
This PR makes the following changes:

* If HTTP/2 is possible, we should use it — that way we can multiplex multiple in-flight HTTP requests onto the same HTTP/2 connection rather than having to open multiple connections
* If connection keepalives are enabled, limit to 1 idle request and set the timeout to 5 minutes
* If connection keepalives are disabled:
  * If HTTP/1 is in use, it will be closed as soon as the request completes
  * If HTTP/2 is in use, it will be closed "as soon as possible" (see golang/net#48)
* Reap unused federation tripper entries for remote hosts after 5 minutes, otherwise we'll just hold onto these things in memory indefinitely even if we stop talking to a given host
* Use a TLS client session cache, which should allow TLS resumption in some cases, potentially saving us some cycles on TLS handshakes

### DIFF
--- a/client.go
+++ b/client.go
@@ -171,15 +171,17 @@ func (f *federationTripper) getTransport(tlsServerName string) (transport http.R
 	// Create the transport if we don't have any for this TLS server name.
 	if transport, ok = f.transports[tlsServerName]; !ok {
 		tr := &http.Transport{
-			DisableKeepAlives: !f.keepAlives,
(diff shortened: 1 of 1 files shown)


kegsay (approved):
It looks alright, but I'd like to see some tests.

PR #298, opened 2022-03-17
Remove `event_id` field from `eventFormatV2Fields`

neilalexander:
This removes `EventID` from `eventFormatV2Fields` and puts a central `eventID` field in the `Event` struct which will either be populated from the `eventFormatV1Fields` (room versions 1-2) or from computing our own event ID (room versions 3 and later).

Beforehand we were trying to find and remove the `event_id` field on every event in room version 3 and later, even though it probably isn't present. The `sjson` call had a few allocations even though it was not really accomplishing anything.

Even if the `event_id` field is present in those room versions, which it really shouldn't be, removing it would break the event signatures, when instead we should just be ignoring what the event ID is and computing our own.

### DIFF
--- a/event.go
+++ b/event.go
@@ -116,13 +116,13 @@ func (eb *EventBuilder) SetUnsigned(unsigned interface{}) (err error) {
 // eventFormatV1Fields, eventFormatV2Fields.
 type Event struct {
 	redacted    bool
+	eventID     string
 	eventJSON   []byte
 	fields      interface{}
 	roomVersion RoomVersion
 }
 
 type eventFields struct {
-	EventID        string     `json:"event_id,omitempty"`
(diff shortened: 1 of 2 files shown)


event.go
```diff
@@ -428,12 +429,10 @@ func (e *Event) populateFieldsFromJSON(eventIDIfKnown string, eventJSON []byte)
 		// Populate the fields of the received object.
 		fields.fixNilSlices()
 		e.fields = fields
+		e.eventID = fields.EventID
 	case EventFormatV2:
 		// Later room versions don't have the event_id field so if it is
 		// present, remove it.
```
kegsay, line 435: Need to update comments.

kegsay (approved):

PR #300, opened 2022-04-05
`RespState` and untrusted event tweaks

neilalexander:
This PR makes the following changes:

* `RespState.Check()` will now return the unmarshalled events as well as modifying the `RespState` JSONs — we go through the effort of unmarshalling so we'd might as well — this will remove an unnecessary marshalling round-trip step when processing `RespSendJoin`;
* Untrusted events now strip the `"unsigned"` key.

### DIFF
--- a/event.go
+++ b/event.go
@@ -318,6 +318,9 @@ func NewEventFromUntrustedJSON(eventJSON []byte, roomVersion RoomVersion) (resul
 		return
 	}
 
+	if eventJSON, err = sjson.DeleteBytes(eventJSON, "unsigned"); err != nil {
+		return
+	}
 	if eventFormat == EventFormatV2 {
 		if eventJSON, err = sjson.DeleteBytes(eventJSON, "event_id"); err != nil {
 			return
--- a/federationtypes.go
+++ b/federationtypes.go
@@ -311,15 +311,14 @@ func (r RespState) Events(roomVersion RoomVersion) []*Event {
 // that do not have valid signatures, and also returns the unmarshalled
 // auth events (first return parameter) and state events (second
 // return parameter).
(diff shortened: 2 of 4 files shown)


event.go
```diff
@@ -318,6 +318,9 @@ func NewEventFromUntrustedJSON(eventJSON []byte, roomVersion RoomVersion) (resul
 		return
 	}
 
+	if eventJSON, err = sjson.DeleteBytes(eventJSON, "unsigned"); err != nil {
```
kegsay, line 321: Won't this nuke `invite_state` etc?

federationtypes.go
```diff
@@ -308,30 +308,35 @@ func (r RespState) Events(roomVersion RoomVersion) []*Event {
 
 // Check that a response to /state is valid. This function mutates
 // the RespState to remove any events from AuthEvents or StateEvents
-// that do not have valid signatures.
-func (r *RespState) Check(ctx context.Context, roomVersion RoomVersion, keyRing JSONVerifier, missingAuth AuthChainProvider) error {
+// that do not have valid signatures, and also returns the unmarshalled
+// auth events (first return parameter) and state events (second
+// return parameter).
+func (r *RespState) Check(ctx context.Context, roomVersion RoomVersion, keyRing JSONVerifier, missingAuth AuthChainProvider) (authEvents, stateEvents []*Event, err error) {
```
kegsay, line 314: Please don't use named returns here, this function is far too long for that

kegsay (approved):

event.go
neilalexander: It doesn't — the stripped state is sent in the invite federation request and then we add it to the `"unsigned"` section ourselves.

PR #302, opened 2022-04-08
Make it possible to correctly parse filters

S7evinK:
This should allow Dendrite to parse filters like ` { "room": "state": { "types": [] } }` correctly.

### DIFF
--- a/filter.go
+++ b/filter.go
@@ -31,17 +31,17 @@ type Filter struct {
 
 // EventFilter is used to define filtering rules for events
 type EventFilter struct {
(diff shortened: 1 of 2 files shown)


neilalexander (changes requested):
I suppose this means we now need to be much more careful checking for `nil`s on the Dendrite side?

neilalexander (approved):
Er, that was supposed to be an approve.

S7evinK:
That's correct, the PR for Dendrite does exactly that.

PR #303, opened 2022-05-05
Rename `federationTripper` to `destinationTripper`

neilalexander:
... since it isn't just used for federation after all.

### DIFF
--- a/client.go
+++ b/client.go
@@ -38,8 +38,9 @@ import (
 // Default HTTPS request timeout
 const requestTimeout time.Duration = time.Duration(30) * time.Second
 
-// A Client makes request to the federation listeners of matrix
-// homeservers
+// A Client makes HTTP requests to remote servers. It is used to
+// centralise a number of configurable options, such as DNS caching,
+// timeouts etc.
 type Client struct {
 	client    http.Client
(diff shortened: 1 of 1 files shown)


client.go
```diff
@@ -131,55 +132,55 @@ func WithKeepAlives(keepAlives bool) ClientOption {
...
-const federationTripperReapInterval = time.Minute // how often to check for dead entries
+const destinationTripperLifetime = time.Minute * 5 // how long to keep an entry
+const destinationTripperReapInterval = time.Minute // how often to check for dead entries
 
 // nolint:maligned
-type federationTripper struct {
+type destinationTripper struct {
 	// transports maps an TLS server name with an HTTP transport.
-	transports      map[string]*federationTripperTransport
+	transports      map[string]*destinationTripperTransport
```
kegsay, line 141: Why do we do this? Why can't we share transports for different endpoints?

kegsay (approved):
Thank you!

client.go
neilalexander: Because we have to set `ServerName` in the `tls.Config` for the remote server SNI, so each destination needs its own `tls.Config`.

PR #304, opened 2022-05-05
Control whether `Client` performs well-known or SRV lookups, update `FederationClient`

neilalexander:
That way we can use `Client` to do regular requests without performing well-known or SRV lookups.

### DIFF
--- a/client.go
+++ b/client.go
@@ -52,11 +52,12 @@ type UserInfo struct {
 }
 
 type clientOptions struct {
-	transport  http.RoundTripper
-	dnsCache   *DNSCache
-	timeout    time.Duration
-	skipVerify bool
-	keepAlives bool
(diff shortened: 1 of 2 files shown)


client.go
```diff
@@ -132,6 +134,13 @@ func WithKeepAlives(keepAlives bool) ClientOption {
 	}
 }
 
+// WithWellKnownSRVLookups enables federation lookups of well-known and SRV records.
+func WithWellKnownSRVLookups(wellKnownSRV bool) ClientOption {
```
S7evinK, line 138: Doesn't `WithWellKnownSRVLookups` already imply `true`?

neilalexander: You could also supply `WithWellKnownSRVLookups(false)` to force them to be disabled, so no.

PR #305, opened 2022-05-06
Produce more useful event auth errors

neilalexander:
The errors from event auth have traditionally been notoriously un-useful. This PR fixes that by making more of an effort to explain why something was rejected.

### DIFF
--- a/eventauth.go
+++ b/eventauth.go
@@ -1103,11 +1103,12 @@ func (m *membershipAllower) membershipAllowedSelf() error { // nolint: gocyclo
 	if m.oldMember.Membership == Leave && m.newMember.Membership == Leave {
 		return nil
 	}
-	if m.newMember.Membership == Knock {
+
+	switch m.newMember.Membership {
+	case Knock:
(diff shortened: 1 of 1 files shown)


eventauth.go
```diff
@@ -1170,8 +1172,15 @@ func (m *membershipAllower) membershipAllowedSelf() error { // nolint: gocyclo
 		if m.oldMember.Membership == Invite {
 			return nil
 		}
+		return m.membershipFailed(
+			"sender cannot leave from this state",
+		)
+
+	default:
+		return m.membershipFailed(
+			"sender cannot set their own membership to %q", m.newMember.Membership,
```
kegsay, line 1181: State that the membership type is unknown?

kegsay (approved):

eventauth.go
neilalexander: In some cases, like `Invite` and `Ban`, it isn't really unknown.

neilalexander: I suppose I can add special cases for those.

PR #308, opened 2022-05-12
Pull in related events into the conflicted control events

neilalexander:
This PR should fix some state res bugs by not allowing an event directly related to a power event from the conflicted set from being reapplied later as a non-power event.


erikjohnston:
Yeah, I think that looks right

PR #321, opened 2022-07-08
Use url.QueryEscape instead of url.PathEscape

S7evinK:
`sed -i 's/PathEscape/QueryEscape/g' federationclient.go`

`url.PathEscape` doesn't escape `@!#` etc, but `url.QueryEscape` does, so use that.


neilalexander:
FYI we’ve been here before and bounced back and forth on this numerous times already, and I think we ran into problems in the past where using `QueryEscape` in the path segments could sometimes break the federation request signing.

PR #322, opened 2022-07-11
Redaction tweaks

neilalexander:
This makes two changes:

1. `(*Event).Redact()` now doesn't return a copy of the event, but instead modifies the event from the receiver, since all we ever seem to do is `event = event.Redact()`;
2. Exports `RedactEventJSON` so that it can be used in combination with `VerifyJSON` without having to round-trip an `*Event`

### DIFF
--- a/event.go
+++ b/event.go
@@ -540,6 +540,7 @@ func (e *Event) Redact() {
 	if err = e.populateFieldsFromJSON(e.EventID(), eventJSON); err != nil {
 		panic(fmt.Errorf("gomatrixserverlib: populateFieldsFromJSON failed %v", err))
 	}
+	e.redacted = true
 }
 
 // SetUnsigned sets the unsigned key of the event.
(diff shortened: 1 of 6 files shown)


event.go
```diff
@@ -537,16 +537,10 @@ func (e *Event) Redact() *Event {
 		// This is unreachable for events created with EventBuilder.Build or NewEventFromUntrustedJSON
 		panic(fmt.Errorf("gomatrixserverlib: invalid event %v", err))
 	}
-	result := Event{
-		redacted:    true,
-		roomVersion: e.roomVersion,
-		eventJSON:   eventJSON,
```
kegsay, line 543: We don't seem to set `eventJSON` to the redacted version?

kegsay (approved):

event.go
neilalexander: Yep, it's done by `populateFieldsFromJSON`.

PR #325, opened 2022-07-14
validate senders and rooms IDs in filters

watson28:
Signed-off-by: Jorge Florian <jafn28@gmail.com>

### Pull Request Checklist

* [ ] Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/master/docs/CONTRIBUTING.md#sign-off)

Signed-off-by: `Jorge Florian <jafn28@gmail.com>`

This PR fixes: https://github.com/matrix-org/dendrite/issues/2067

### DIFF
--- a/eventcontent.go
+++ b/eventcontent.go
@@ -19,9 +19,32 @@ import (
 	"database/sql/driver"
 	"encoding/json"
 	"fmt"
+	"regexp"
 	"strconv"
 	"strings"
 )
+
+var userIDRegexp *regexp.Regexp
+var roomIDRegexp *regexp.Regexp
+
+func init() {
+	// https://spec.matrix.org/v1.3/appendices/#server-name
+	IPV4AddressExpr := `\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}`
+	IPV6AddressExpr := `\[[\d\x41-\x46\x61-\x66\:\.]{2,45}\]`
+	dnsNameExpr := `[\w\-\.]{1,255}`
+	portExpr := `\:\d{1,5}`
+	servernameExpr := fmt.Sprintf(`(%s|%s|%s)(%s)?`, IPV4AddressExpr, IPV6AddressExpr, dnsNameExpr, portExpr)
+
+	// Regexp from https://spec.matrix.org/v1.3/appendices/#user-identifiers
+	// and https://spec.matrix.org/v1.3/appendices/#historical-user-ids
+	localpartExpr := `[\x21-\x39\x3B-\x7E]+`
+
+	userIDExpr := fmt.Sprint(`^@`, localpartExpr, `\:`, servernameExpr, `$`)
+	userIDRegexp = regexp.MustCompile(userIDExpr)
+
+	roomIDExpr := fmt.Sprint(`^!`, localpartExpr, `\:`, servernameExpr, `$`)
+	roomIDRegexp = regexp.MustCompile(roomIDExpr)
+}
 
 // CreateContent is the JSON content of a m.room.create event along with
 // the top level keys needed for auth.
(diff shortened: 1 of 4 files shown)


eventcontent.go
```diff
@@ -519,6 +542,10 @@ func (v *levelJSONValue) assignIfExists(to *int64) {
...
 // Check if the user ID is a valid user ID.
 func isValidUserID(userID string) bool {
-	// TODO: Do we want to add anymore checks beyond checking the sigil and that it has a domain part?
-	return userID[0] == '@' && strings.IndexByte(userID, ':') != -1
+	return userIDRegexp.MatchString(userID)
+}
+
+// Check if the room ID is a valid room ID.
+func isValidRoomID(roomID string) bool {
+	return roomIDRegexp.MatchString((roomID))
```
neilalexander, line 550: `(roomID)` should be sufficient here instead of `((roomID))`.

eventcontent.go
```diff
@@ -19,10 +19,33 @@ import (
...
+var userIDRegexp *regexp.Regexp
+var roomIDRegexp *regexp.Regexp
+
+func init() {
+	// https://spec.matrix.org/v1.3/appendices/#server-name
+	IPV4AddressExpr := `\d{1,3}\.\d{1,3}.\d{1,3}.\d{1,3}`
+	IPV6AddressExpr := `[\d\x41-\x46\x61-\x66\:\.]{2,45}`
+	dnsNameExpr := `[\w\-\.]{1,255}`
+	portExpr := `\:\d{1,5}`
+	servernameExpr := fmt.Sprint(`(`, IPV4AddressExpr, `|[`, IPV6AddressExpr, `]|`, dnsNameExpr, `)(`, portExpr, `)?`)
```
neilalexander, line 36: For clarity can we use `fmt.Sprintf` here with a format string, like `fmt.Sprintf(`%s, %s, %s, %s`, ...)`?

Also is there a reason why `IPV6AddressExpr` is surrounded in more square brackets?

filter.go
```diff
@@ -82,6 +82,28 @@ func (filter *Filter) Validate() error {
...
+		return result
+	}
+	return nil
+}
+
+func (filter *RoomEventFilter) Validate() error {
+	if filter.Rooms != nil {
+		for _, roomID := range *filter.Rooms {
+			if !isValidRoomID(roomID) {
+				return errors.New("Bad room value. Must be in the form !localpart:domain")
```
neilalexander, line 95: It would be useful if the error here can return the faulty value, i.e. `return fmt.Errorf("Bad room value %q: must be in form...", roomID)`

eventcontent.go
```diff
@@ -519,6 +542,10 @@ func (v *levelJSONValue) assignIfExists(to *int64) {
 
 // Check if the user ID is a valid user ID.
 func isValidUserID(userID string) bool {
-	// TODO: Do we want to add anymore checks beyond checking the sigil and that it has a domain part?
-	return userID[0] == '@' && strings.IndexByte(userID, ':') != -1
+	return userIDRegexp.MatchString(userID)
```
tulir, line 545: `isValidUserID` seems to be used in [power level event auth](https://github.com/matrix-org/gomatrixserverlib/blob/5ff38e2c2839cc21cc37d9e5fbaf85386ffdf5f4/eventauth.go#L512-L518), which means this is not safe (at least not before MSC2828). Right now it's possible to make user and room IDs with arbitrary unicode (spaces, emojis, everything). For example, `@🐈️:maunium.net` can be found in #dendrite-dev:matrix.org

I think filters aren't used for anything critical, but obviously validating them means you won't be able to filter user/room IDs containing non-ascii characters.

eventcontent.go
watson28: Yes, that's part of the grammar specified at https://spec.matrix.org/v1.3/appendices/#server-name.
IP v6 addresses should be wrapped in square brackets when there is a prefix or suffix as part of a URI.

I forgot to scape those square brackets, so they are not interpreted as Regex character classes. But now it's fixed.

PR #327, opened 2022-08-01
Preserve history visibility in `HeaderedReverseTopologicalOrdering`

S7evinK:
This preserves the history visibility when using `HeaderedReverseTopologicalOrdering`, otherwise incremental syncs in Dendrite can't check the visibility of an event.


stateresolutionv2.go
```diff
@@ -231,13 +231,16 @@ func ReverseTopologicalOrdering(input []*Event, order TopologicalOrder) []*Event
 func HeaderedReverseTopologicalOrdering(events []*HeaderedEvent, order TopologicalOrder) []*HeaderedEvent {
 	r := stateResolverV2{}
 	input := make([]*Event, len(events))
+	hisVis := make(map[string]HistoryVisibility)
```
neilalexander, line 234: We know the capacity here ahead of time as `len(events)` so please give that to `make()` to reduce the number of allocations.

neilalexander (approved):

PR #329, opened 2022-08-11
Add a `SendKnock` method and `RespSendKnock` struct

reivilibre:
I also need these for writing a Complement test.

Q: is `EventJSONs` the right type to use for `knock_room_state`?
The spec describes it as `[StrippedState]`. `EventJSONs` seems to be an alias for an array of `RawJSON`, so sounds like it should be compatible at the very least.


neilalexander:
> Q: is EventJSONs the right type to use for knock_room_state?

If you just plan to hold JSON and not do anything with it then yes, but if you want to turn them into actual `*Event` then no, stripped events won't work here.

reivilibre:
> > Q: is EventJSONs the right type to use for knock_room_state?
> 
> If you just plan to hold JSON and not do anything with it then yes, but if you want to turn them into actual `*Event` then no, stripped events won't work here.

I believe that's what I want then, but let me know if that's not what you expect.

federationtypes.go
```diff
@@ -448,6 +448,12 @@ func (r RespSendJoin) MarshalJSON() ([]byte, error) {
 	return json.Marshal(partialJoinFields)
 }
 
+// A RespSendKnock is the content of a response to PUT /_matrix/federation/v2/send_knock/{roomID}/{eventID}
+type RespSendKnock struct {
+	// A list of stripped state events to help the initiator of the knock identify the room.
+	KnockRoomState EventJSONs `json:"knock_room_state"`
```
neilalexander, line 454: Thinking some more about this, the fields from the spec are `content`, `state_key`, `sender` and `type` which is the exact same shape as the `InviteV2StrippedState` type. If you want to be able to access those fields without having to run through the JSON unmarshaller in your tests, then it might be a better idea to define this as:
```
KnockRoomState []InviteV2StrippedState `json:"knock_room_state"`
```

neilalexander (approved):
Overall looks OK!

PR #330, opened 2022-08-18
Add a MakeKnock federation client function

reivilibre:
I have also corrected the field name in the response to `KnockEvent`; of course I'd forget to update that instance of 'join' :facepalm:.


federationclient.go
```diff
@@ -68,6 +68,21 @@ func (ac *FederationClient) SendTransaction(
 	return
 }
 
+// Creates a version query string with all the specified room versions, typically
+// the list of all supported room versions.
+// Needed when making a /make_knock or /make_join request.
+func makeVersionQueryString(roomVersions []RoomVersion) string {
+	versionQueryString := ""
+	if len(roomVersions) > 0 {
+		var vqs []string
```
neilalexander, line 77: Just as an FYI, if you do `vqs := make([]string, 0, len(roomVersions))` instead to set the initial capacity to the same length as the supplied `[]RoomVersion` slice, then you can avoid reallocations if you append more items than the default initial capacity. It's a good little optimisation to get used to in cases that you already know how many items you will probably have.

neilalexander (approved):

PR #335, opened 2022-09-05
State/auth tweaks

neilalexander:
This PR makes a couple minor changes:

* refactors the create check slightly, also checks that the room version is valid 
* modifies the state res v2 `applyEvents` function so that it doesn't accidentally drop events with unexpected non-empty state keys (i.e. `{"m.room.create", "something"}`) into oblivion, but instead sorts them into "other" events, so that event auth can decide the right thing to do with them instead
* modifies the state res v2 `isControlEvent` function to also check state keys, so that we don't incorrectly classify something that shouldn't be a control event


kegsay (approved):
+100 much better control flow on `applyEvents`, thanks!

PR #336, opened 2022-09-06
Update `applyAndAuthEvents`

neilalexander:
This updates `applyAndAuthEvents` in state res v2 to pick auth events from the event if there is no matching event type/state key tuple in the partial state. It also optimises the auth event provider so that it only unmarshals the contents of events if they have changed to save CPU/allocations.

### DIFF
--- a/stateresolutionv2.go
+++ b/stateresolutionv2.go
@@ -155,13 +155,6 @@ func ResolveStateConflictsV2(
 		}
 	}
 
-	// Before we can do anything with the conflicted and unconflicted sets, we
-	// first need to auth and apply the entire auth chain in order. This is so that
-	// when we come to auth any future events against the partial state, we'll have
-	// the knowledge from the auth chain to help us to make a correct decision.
-	authEvents = r.reverseTopologicalOrdering(authEvents, TopologicalOrderByAuthEvents)
-	r.authAndApplyEvents(authEvents)
-
 	// Then process the unconflicted events by ordering them topologically and then
 	// authing them. The successfully authed events will form the real initial partial
 	// state. We will then keep the successfully authed unconflicted events so that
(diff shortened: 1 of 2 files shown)


stateresolutionv2.go
```diff
@@ -155,13 +155,6 @@ func ResolveStateConflictsV2(
 		}
 	}
 
-	// Before we can do anything with the conflicted and unconflicted sets, we
-	// first need to auth and apply the entire auth chain in order. This is so that
-	// when we come to auth any future events against the partial state, we'll have
-	// the knowledge from the auth chain to help us to make a correct decision.
-	authEvents = r.reverseTopologicalOrdering(authEvents, TopologicalOrderByAuthEvents)
-	r.authAndApplyEvents(authEvents)
```
kegsay, line 163: Why don't we do this anymore?

stateresolutionv2.go
```diff
@@ -387,21 +380,39 @@ func (r *stateResolverV2) getFirstPowerLevelMainlineEvent(event *Event) (
 // the event is ignored and dropped. Returns two lists - the first contains the
 // accepted (authed) events and the second contains the rejected events.
 func (r *stateResolverV2) authAndApplyEvents(events []*Event) {
+	authEvents := NewAuthEvents(nil)
 	for _, event := range events {
-		// Check if the event is allowed based on the current partial state. If the
-		// event isn't allowed then simply ignore it and process the next one.
+		// Collect together the auth events. We'll start by collecting the
+		// auth event IDs from the event itself. These are our fallback if
+		// no partial state events are known.
```
kegsay, line 387: Is this safe? Does synapse do this?

stateresolutionv2.go
neilalexander: It's no longer necessary, since we can pick out from the auth events in `authAndApplyEvent` if needed. I think this actually was a bug that resulted in an incorrect partial state up front.

stateresolutionv2.go
neilalexander: Yes, from the spec:

> Iterative auth checks. ... If a (event_type, state_key) key that is required for checking the authorization rules is not present in the state, then the appropriate state event from the event’s auth_events is used if the auth event is not rejected.

stateresolutionv2.go
```diff
@@ -387,21 +380,39 @@ func (r *stateResolverV2) getFirstPowerLevelMainlineEvent(event *Event) (
...
+			if event != nil {
+				_ = authEvents.AddEvent(event)
+			}
+		}
+
+		// Check if the event is allowed based on the current partial state.
+		r.allower.update(&authEvents)
 		if err := r.allower.allowed(event); err != nil {
+			// The event was not allowed by the partial state, so skip it.
 			continue
```
kegsay, line 410: Partial state *and* fallback auth events presumably?

kegsay (approved):

stateresolutionv2.go
neilalexander: That's right, yes, I'll update the comment.

PR #337, opened 2022-09-07
Add GHA workflow

S7evinK:
Add GHA workflow that runs
- `golangci-lint`
- `go test -race ./...` on GMSL
- `go test ./...` on Dendrite with the new GMSL version

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,0 +1,107 @@
+name: Tests
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  # Run golangci-lint
+  lint:
+    timeout-minutes: 5
+    name: Linting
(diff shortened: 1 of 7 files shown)


neilalexander (approved):
Good idea!

.github/workflows/tests.yml
```diff
@@ -0,0 +1,105 @@
...
+          go-version: ${{ matrix.go }}
+      - uses: actions/cache@v3
+        with:
+          path: |
+            ~/.cache/go-build
+            ~/go/pkg/mod
+          key: ${{ runner.os }}-go${{ matrix.go }}-test-${{ hashFiles('**/go.sum') }}
+          restore-keys: |
+            ${{ runner.os }}-go${{ matrix.go }}-test-
+      - run: go get github.com/matrix-org/gomatrixserverlib@${{ github.event.pull_request.head.sha }} && go mod tidy
```
neilalexander, line 98: I'm somewhat wondering what this will do in a PR scenario from outside the repository. Maybe we can bind-mount in the checked out source instead?

PR #338, opened 2022-09-09
Enforce `type`, `sender`, `state_key` and `room_id` lengths using bytes rather than codepoints

neilalexander:
This effectively reverts the change made in [5f66df0](https://github.com/matrix-org/gomatrixserverlib/commit/5f66df0) to bytes instead of codepoints, since Synapse will now enforce the same after matrix-org/synapse#13710. 

History here is that Synapse originally calculated bytes in Python 2.x, started counting codepoints in Python 3.x pretty much by accident and then the spec was ambiguous after the fact (hence matrix-org/matrix-spec#1001).

Rationale is that bytes are probably easier for implementations to manage and less likely to generate huge indexes for client-side databases (especially where limits might exist like LMDB). 

cc @reivilibre

### DIFF
--- a/event.go
+++ b/event.go
@@ -728,18 +728,18 @@ func (e *Event) CheckFields() error { // nolint: gocyclo
 		}
 	}
 
-	if l := len([]rune(fields.Type)); l > maxIDLength {
+	if l := len(fields.Type); l > maxIDLength {
 		return EventValidationError{
 			Code:    EventValidationTooLarge,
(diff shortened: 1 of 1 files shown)


DMRobertson:
> History here is that Synapse originally calculated bytes in Python 2.x,

We later realised this is not the case: see Olivier's writeup in https://github.com/matrix-org/matrix-spec/issues/1001#issuecomment-1241872952

PR #343, opened 2022-10-12
Validate auth events are from the same room

S7evinK:
for https://github.com/matrix-org/dendrite/pull/2791


eventauth.go
```diff
@@ -313,6 +321,12 @@ func (a *AuthEvents) AddEvent(event *Event) error {
 	if event.StateKey() == nil {
 		return fmt.Errorf("AddEvent: event %q does not have a state key", event.Type())
 	}
+	for _, ev := range a.events {
+		if ev.RoomID() != "" && ev.RoomID() != event.RoomID() {
```
neilalexander, line 325: Is `ev.RoomID() != ""` ever going to be true?

S7evinK: Ah, a relic from other tests before. :D

PR #350, opened 2023-01-12
Use stable identifiers from MSC3706 (Faster joins)

DMRobertson:
I haven't made any effort to maintain backwards compatibility (unlike in https://github.com/matrix-org/synapse/pull/14832). I'm happy to do so if needed, but might need some guidance.

Synapse issue: https://github.com/matrix-org/synapse/issues/14828
Synapse PR: https://github.com/matrix-org/synapse/pull/14832
Complement PR: https://github.com/matrix-org/complement/pull/583
GMSL PR: https://github.com/matrix-org/gomatrixserverlib/pull/350
MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/3706
Spec diff: https://github.com/matrix-org/matrix-spec/pull/1393

### Pull Request Checklist

Signed-off-by: `David Robertson <davidr@element.io>`

### DIFF
--- a/federationtypes.go
+++ b/federationtypes.go
@@ -444,9 +444,9 @@ type RespSendJoin struct {
 	// but not guaranteed to be present as it's only since MSC3083.
 	Event RawJSON `json:"event,omitempty"`
 	// true if the state is incomplete
-	PartialState bool `json:"org.matrix.msc3706.partial_state"`
+	PartialState bool `json:"members_omitted"`
 	// a list of servers in the room. Only returned if partial_state is set.
-	ServersInRoom []string `json:"org.matrix.msc3706.servers_in_room"`
+	ServersInRoom []string `json:"servers_in_room"`
 }
 
 // MarshalJSON implements json.Marshaller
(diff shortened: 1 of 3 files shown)


federationtypes.go
```diff
@@ -444,9 +444,9 @@ type RespSendJoin struct {
 	// but not guaranteed to be present as it's only since MSC3083.
 	Event RawJSON `json:"event,omitempty"`
 	// true if the state is incomplete
-	PartialState bool `json:"org.matrix.msc3706.partial_state"`
+	PartialState bool `json:"members_omitted"`
```
richvdh, line 447: I kinda wonder if we should rename the struct field while we're here. It'll break compatibility with any apps using it, but there doesn't seem to be any advantage to delaying it?

DMRobertson: Not sure what the done thing is in GMSL, but I'm persuaded.

DMRobertson: See d5bef8c

kegsay (approved):
Don't worry about backwards compat - we don't use faster joins in Dendrite yet.

PR #351, opened 2023-01-19
Fix `checkUserLevels` and add tests for it

S7evinK:
Fixes the discussed issue with user levels

### DIFF
--- a/eventauth.go
+++ b/eventauth.go
@@ -700,6 +700,14 @@ func checkUserLevels(senderLevel int64, senderID string, oldPowerLevels, newPowe
 		}
 	}
 
+	// also add old levels to check for e.g. deletions
+	for userID := range oldPowerLevels.Users {
+		userLevelChecks[userID] = levelPair{
+			old: oldPowerLevels.UserLevel(userID),
+			new: newPowerLevels.UserLevel(userID),
+		}
+	}
+
 	// Check each of the levels in the list.
 	for userID, level := range userLevelChecks {
 		// Check if the level is being changed.
--- a/eventauth.go
+++ b/eventauth.go
@@ -935,7 +935,7 @@ func (e *eventAllower) commonChecks(event *Event) error {
 	eventLevel := e.powerLevels.EventLevel(event.Type(), stateKey != nil)
 	if senderLevel < eventLevel {
 		return errorf(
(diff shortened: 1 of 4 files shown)


eventauth.go
```diff
@@ -927,7 +935,7 @@ func (e *eventAllower) commonChecks(event *Event) error {
 	eventLevel := e.powerLevels.EventLevel(event.Type(), stateKey != nil)
 	if senderLevel < eventLevel {
 		return errorf(
-			"sender %q is not allowed to send event. %d < %d",
+			"sender %q is not allowed to send powerlevel event. %d < %d",
```
kegsay, line 938: How do we know this is only for PL events?

eventauth_test.go
```diff
@@ -1200,3 +1200,201 @@ func TestNegativePowerLevels(t *testing.T) {
...
+				},
+				newPowerLevels: PowerLevelContent{
+					Users: map[string]int64{
+						senderID: 100,
+					},
+				},
+			},
+		},
+		{
+			name: "removing other user level if NOT equal our own is allowed",
```
kegsay, line 1237: "NOT equal" isn't precise enough, please ensure you test > and <.

eventauth_test.go
```diff
@@ -1200,3 +1200,201 @@ func TestNegativePowerLevels(t *testing.T) {
...
+				},
+				newPowerLevels: PowerLevelContent{
+					Users: map[string]int64{
+						senderID: 101,
+					},
+				},
+			},
+		},
+		{
+			name: "setting own user level below own is allowed",
```
kegsay, line 1309: Also check setting level to the same level please.

eventauth_test.go
```diff
@@ -1200,3 +1200,201 @@ func TestNegativePowerLevels(t *testing.T) {
...
+						senderID:    100,
+						"@bob:test": 110,
+					},
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := checkUserLevels(tt.args.senderLevel, senderID, tt.args.oldPowerLevels, tt.args.newPowerLevels); (err != nil) != tt.wantErr {
```
kegsay, line 1393: Break this up please, it's too confusing to read.

kegsay (approved):

PR #359, opened 2023-04-06
Remove .Check() functions; migrate them to GMSL static funcs

kegsay:
For now, pending review.


devonh (approved):
Seems good. Is any of this tested?

kegsay:
No, as it's just migrating from struct methods to static functions. It's indirectly tested by dendrite tests passing on the dendrite PR.

PR #365, opened 2023-04-20
Invert public API to call methods on the RoomVersion

kegsay:
AKA Changes `NewEventFromTrustedJSON(event, RoomVersionV3)` into `RoomVersionV3.NewEventFromTrustedJSON(event)`. Remove other event creation public functions. This does 2 things:
 - Funnels event creation into fewer functions.
 - Allows us to swap `type RoomVersion string` with `type RoomVersion interface { NewEventFromTrustedJSON(...) }` later on.


devonh (approved):
looks good

PR #366, opened 2023-04-20
Pull fedclient interface over from dendrite

devonh:


fclient/federationclient.go
```diff
@@ -14,9 +15,60 @@ import (
 	"golang.org/x/crypto/ed25519"
 )
 
+// an interface for gmsl.FederationClient - contains functions called by federationapi only.
+type FederationClient interface {
+	gomatrixserverlib.KeyClient
+
+	SetInternalClient(client Client)
+	WithUserAgent(agent string)
```
kegsay, line 23: I meant to use this as a `ClientOption` for use with `NewFederationClient`

fclient/federationclient.go
```diff
@@ -14,9 +15,60 @@ import (
 	"golang.org/x/crypto/ed25519"
 )
 
+// an interface for gmsl.FederationClient - contains functions called by federationapi only.
+type FederationClient interface {
+	gomatrixserverlib.KeyClient
+
+	SetInternalClient(client Client)
```
kegsay, line 22: We want to get rid of this I thought?

fclient/federationclient.go
```diff
@@ -14,9 +15,60 @@ import (
...
+	WithUserAgent(agent string)
+	DoRequestAndParseResponse(ctx context.Context, req *http.Request, result interface{}) error
```
kegsay, line 24: We'll probably need the `CreateMediaDownloadRequest` or whatever it is from `Client` being brought up to this interface. Can we rename it to `DownloadMedia` please?

fclient/federationclient_test.go
```diff
@@ -58,7 +58,7 @@ func TestSendJoinFallback(t *testing.T) {
 		},
 		fclient.WithSkipVerify(true),
 	)
-	fc.Client = *fclient.NewClient(fclient.WithTransport(
+	fc.SetInternalClient(*fclient.NewClient(fclient.WithTransport(
```
kegsay, line 61: We should be able to use `fc.NewFederationClient(signingIdentity, fclient.WithTransport(...))` instead of gut wrenching the client?

kegsay (approved):

fclient/federationclient.go
devonh: ah I see

PR #371, opened 2023-04-24
refactor: create a PDU interface

kegsay:
Only one type assertion, which is in `(e *HeaderedEvent) UnmarshalJSON` as that then nests the *Event.


kegsay:
Closing this as https://github.com/matrix-org/gomatrixserverlib/pull/373 needs to land first.

PR #372, opened 2023-04-24
Bring high level room joining logic into GMSL

devonh:


fclient/performjoin.go
```diff
@@ -0,0 +1,238 @@
...
+		ctx,
+		origin,
+		input.ServerName,
+		input.RoomID,
+		input.UserID.Raw(),
+	)
+	if err != nil {
+		// TODO: Check if the user was not allowed to join the room.
+		callbacks.FederationFailure(input.ServerName)
+		return nil, nil, fmt.Errorf("r.federation.MakeJoin: %w", err)
```
kegsay, line 57: The use of callbacks here feels weird. I'd expect us to have a defined errror type e.g `FederationFailureError` which we then return, which callers can then type-switch against.

fclient/performjoin.go
```diff
@@ -0,0 +1,238 @@
...
+	}
+	callbacks.FederationSuccess(input.ServerName)
```
kegsay, line 59: In practice what do you expect callers to do when they get this callback? If it's just for tracking blacklist status, I would really just return a flag to indicate if the destination server was reachable.

fclient/performjoin.go
```diff
@@ -0,0 +1,238 @@
...
+	// contain signatures that we don't know about.
+	if len(respSendJoin.Event) > 0 {
+		var remoteEvent *gomatrixserverlib.Event
+		remoteEvent, err = verImpl.NewEventFromUntrustedJSON(respSendJoin.Event)
+		if err == nil && isWellFormedMembershipEvent(
+			remoteEvent, input.RoomID, input.UserID,
+		) {
+			event = remoteEvent
+		}
+	}
```
kegsay, line 130: We seemingly continue blindly if the event is invalid?

fclient/performjoin.go
```diff
@@ -0,0 +1,238 @@
...
+
+	// Sanity-check the join response to ensure that it has a create
+	// event, that the room version is known, etc.
+	authEvents := respSendJoin.AuthEvents.UntrustedEvents(respMakeJoin.RoomVersion)
+	if err = checkEventsContainCreateEvent(authEvents); err != nil {
+		return nil, nil, fmt.Errorf("sanityCheckAuthChain: %w", err)
+	}
+
+	// Process the join response in a goroutine. The idea here is
```
kegsay, line 139: No goroutines here?

fclient/performjoin.go
```diff
@@ -0,0 +1,238 @@
...
+	// that we'll try and wait for as long as possible for the work
+	// to complete, but if the client does give up waiting, we'll
+	// still continue to process the join anyway so that we don't
+	// waste the effort.
```
kegsay, line 143: The key thing that does this is using a background context rather than the provided `ctx`.

fclient/performjoin.go
```diff
@@ -0,0 +1,238 @@
...
+	return gomatrixserverlib.RoomVersionV4
+}
+
+// isWellFormedMembershipEvent returns true if the event looks like a legitimate
+// membership event.
+func isWellFormedMembershipEvent(event *gomatrixserverlib.Event, roomID string, userID *spec.UserID) bool {
+	if membership, err := event.Membership(); err != nil {
+		return false
+	} else if membership != spec.Join {
+		return false
```
kegsay, line 201: This is misleading. The function is `isWellFormedMembershipEvent` _not_ `isWellFormedJoinMemberEvent` so failing the check if the membership != join is deceptive. Please rename this function.

fclient/performjoin.go
```diff
@@ -0,0 +1,238 @@
+package fclient
```
kegsay, line 1: I don't like this file being in `fclient`. The point of `fclient` was to have dumb boring HTTP wrappers, of which this is not. Please move this to the root level and we will shift it to a `/pdu` package when we're ready.

This means you cannot use `FederationClient` else you'll get a circular import error. Instead, define a `JoinClient` interface with the subset of functions from `FederationClient` that you need (in this case, make join and send join).

fclient/performjoin.go
```diff
@@ -0,0 +1,238 @@
...
+	FederationFailure func(serverName spec.ServerName)
+	FederationSuccess func(serverName spec.ServerName)
+}
+
+func PerformJoin(
+	ctx context.Context,
+	fedClient FederationClient,
+	input PerformJoinInput,
+	callbacks PerformJoinCallbacks,
+) (*gomatrixserverlib.HeaderedEvent, gomatrixserverlib.StateResponse, error) {
```
kegsay, line 42: I don't think this function signature is ideal, because it doesn't really handle errors particularly elegantly. The return args are otherwise good (the join event and the state response, decoupled from the format it is over federation). Error wise, there's a few things we care about:
 - Server is unreachable. Transient error.
 - The server's response is invalid. Permanent error.
 - There's a problem on our side e.g unknown room version or failed to sign the event. Permanent error.

In addition, we care if the server was reachable and we received an error, for blacklist purposes. I would probably define a general `FederationError` in GMSL (not fclient as it isn't http specific) which looks something like:
```go
type FederationError struct {
    ServerName spec.ServerName
    Transient bool
    Reachable bool
    Err error
}

func (e FederationError) Error() string {
    return fmt.Sprintf("FederationError(t=%v,r=%v): %w", e.Transient, e.Reachable, e.Err)
}
``` 

This would mean:
 - Server is unreachable. Transient=true, Reachable=false
 - The server's response is invalid. Transient=false, Reachable=true
 - Failure to sign the event on our side. Transient=false, Reachable=true
 - Failure to talk to our own DB e.g to load keys in the KeyRing. Transient=true, Reachable=true

kegsay (changes requested):
Some API changes would be good I think.

fclient/performjoin.go
devonh: I don't know what the logic should be if you receive a bad event inside a send_join response.
Synapse just accepts the event as is and doesn't seem to do any checks against it. So we differ from them in that regard. 

At this point in the join dance, the resident server has already accepted that event into it's room graph and propagated it out to the other servers in the room. So for us to reject it leaves us in an awkward situation.

fclient/federationclient.go
```diff
@@ -191,8 +203,8 @@ func makeVersionQueryString(roomVersions []gomatrixserverlib.RoomVersion) string
 // See https://matrix.org/docs/spec/server_server/unstable.html#joining-rooms
 func (ac *federationClient) MakeJoin(
 	ctx context.Context, origin, s spec.ServerName, roomID, userID string,
-	roomVersions []gomatrixserverlib.RoomVersion,
 ) (res RespMakeJoin, err error) {
+	roomVersions := roomVersionsToList(gomatrixserverlib.StableRoomVersions())
```
kegsay, line 207: By using stable room versions we make it impossible to make_join with unstable room versions e.g MSCs, and there seems to be no way to say "no really, please do".

join.go
```diff
@@ -0,0 +1,26 @@
...
+
+import (
+	"context"
+
+	"github.com/matrix-org/gomatrixserverlib/spec"
+)
+
+type FederatedJoinClient interface {
+	MakeJoin(ctx context.Context, origin, s spec.ServerName, roomID, userID string) (res MakeJoinResponse, err error)
+	SendJoin(ctx context.Context, origin, s spec.ServerName, event *Event) (res SendJoinResponse, err error)
```
kegsay, line 11: This is okay for now, but I'm wondering if we can be less specific long term. If we ever decide to change the flow here (removing make_join, adding a third stage) then everything needs to change which is unfortunate. It would be nice to hide this behind the room version PDU bits longer term I guess?

performjoin.go
```diff
@@ -0,0 +1,276 @@
...
+
+	PrivateKey ed25519.PrivateKey
+	KeyID      KeyID
+	KeyRing    *KeyRing
+
+	EventProvider EventProvider
+}
+
+type PerformJoinResponse struct {
+	JoinEvent     *HeaderedEvent
```
kegsay, line 29: Please use `*Event` as it'll make merge conflicts easier to wrangle for me. You have the same data on `*Event`, so use `JoinEvent.Version()` to get the `RoomVersion`.

performjoin.go
```diff
@@ -0,0 +1,276 @@
...
+		input.ServerName,
+		input.RoomID,
+		input.UserID.Raw(),
+	)
+	if err != nil {
+		// TODO: Check if the user was not allowed to join the room.
+		return nil, &FederationError{
+			ServerName: input.ServerName,
+			Transient:  true,
+			Reachable:  false,
```
kegsay, line 57: This assumes a network error, but I'm fairly sure we will return an `err` on non-2xx as well.

performjoin.go
```diff
@@ -0,0 +1,276 @@
...
+		context.Background(),
+		roomVersion, StateResponse(respSendJoin),
+		input.KeyRing,
+		event,
+		input.EventProvider,
+	)
+	if err != nil {
+		return nil, &FederationError{
+			ServerName: input.ServerName,
+			Transient:  false,
```
kegsay, line 186: Hard to say if this is transient as it may have failed due to failing to get server keys over the network.

kegsay (approved):
Overall LGTM!

kegsay:
Oh and please add tests for the main logic here. It should be easy now as we've interfaced up the fed client!

performjoin.go
devonh: Yup. This keeps the behaviour the same as before the refactor. I tried to limit behaviour changes in this PR

PR #373, opened 2023-04-25
Remove HeaderedEvent

kegsay:
- It had zero extra data beyond the fields in Event
- It complicates moving to a PDU interface as lots of Dendrite
  faffs about converting events to headered events
- The serialisation format is now represented with the combination of
  `event.ToHeaderedJSON() ([]byte, error)` and `NewEventFromHeaderedJSON(headeredEventJSON []byte, redacted bool) (*Event, error)`
  Normal `json.Marshal(event)` does the Right thing and just marshals the event, without
  the custom `_` fields.

With this change, it will now be possible to use a PDU interface
instead of a concrete `Event` type.


fclient/invitev2.go
```diff
@@ -44,6 +45,7 @@ type InviteV2Request struct {
 
 // MarshalJSON implements json.Marshaller
 func (i InviteV2Request) MarshalJSON() ([]byte, error) {
+	fmt.Println(string(i.fields.Event.JSON()))
```
S7evinK, line 48: I guess this can be removed.

S7evinK (approved):
LGTM

PR #374, opened 2023-04-27
Add PDU interface

kegsay:
Only used for event auth for now, which includes state resolution.

Isn't used in fclient yet, as I suspect there will be issues with marshalling/unmarshalling.


S7evinK (approved):

devonh (approved):
Looks great!

PR #379, opened 2023-05-03
Add ProtoEvent and IRoomVersion.NewEventBuilder

kegsay:
A proto event returned from /make_join is not the same as an EventBuilder.


S7evinK (approved):
Seems reasonable

PR #381, opened 2023-05-11
Move MakeJoin logic from Dendrite to GMSL

devonh:


join.go
```diff
@@ -12,6 +12,21 @@ type FederatedJoinClient interface {
 	SendJoin(ctx context.Context, origin, s spec.ServerName, event PDU) (res SendJoinResponse, err error)
 }
 
+type RoomInfo struct {
+	Version RoomVersion
+	NID     int64
```
S7evinK, line 17: I somewhat feel like that a NID shouldn't be in GMSL. It's more of a concept in Dendrite.

handlejoin.go
```diff
@@ -0,0 +1,380 @@
...
+	if input.UserID.Domain() != input.RequestOrigin {
+		return nil, &util.JSONResponse{
+			Code: http.StatusForbidden,
+			JSON: spec.Forbidden("The join must be sent by the server of the user"),
+		}
+	}
+
+	if input.RoomQuerier == nil {
+		util.GetLogger(input.Context).Error("Missing valid RoomQuerier")
+		return nil, &util.JSONResponse{
```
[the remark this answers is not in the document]
devonh: Ya, it'll never correct itself so it will always return an error. Panic makes more sense.

spec/eventtypes.go
```diff
@@ -1,5 +1,7 @@
 package spec
 
+type MatrixEventType string
```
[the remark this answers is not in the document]
devonh: I will remove it from this PR. In the future I would like to differentiate between strings for event type, join rule, membership state, history visibility, etc. Right now it's just a big mix of raw strings with no type enforcement on it.

handlejoin.go
```diff
@@ -0,0 +1,278 @@
...
+	UserID             spec.UserID
+	RoomID             spec.RoomID
+	RoomVersion        RoomVersion
+	RemoteVersions     []RoomVersion
+	RequestOrigin      spec.ServerName
+	RequestDestination spec.ServerName
+	LocalServerName    spec.ServerName
+	LocalServerInRoom  bool
+	RoomQuerier        JoinRoomQuerier
+	BuildEventTemplate func(*ProtoEvent) (PDU, []PDU, error)
```
kegsay, line 23: Needs comments on what the return args are.

handlejoin.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+type HandleMakeJoinResponse struct {
+	JoinTemplateEvent ProtoEvent
+	RoomVersion       RoomVersion
+}
+
+func HandleMakeJoin(input HandleMakeJoinInput) (*HandleMakeJoinResponse, error) {
```
kegsay, line 31: Naming: We're using `HandleMakeJoin` everywhere (this function, Response, and Input). Why not just `JoinRoom`? This is a top-level function, so shouldn't be confused with the `fclient` version?

handlejoin.go
```diff
@@ -0,0 +1,278 @@
...
+	"github.com/matrix-org/gomatrixserverlib/spec"
+	"github.com/matrix-org/util"
+)
+
+type HandleMakeJoinInput struct {
+	Context            context.Context
+	UserID             spec.UserID
+	RoomID             spec.RoomID
+	RoomVersion        RoomVersion
+	RemoteVersions     []RoomVersion
```
kegsay, line 17: Unclear why there are 2 sets of versions here. I guess the `[]` is "I know these versions" but why is it dictating what version the room is to join? How does it even know that info pre make_join?

handlejoin.go
```diff
@@ -0,0 +1,278 @@
...
+	RequestOrigin      spec.ServerName
```
kegsay, line 18: Do we need the request origin or can we just extract this from the `UserID`? It would be a cleaner API as it would stop "known broken" configurations (i.e it is never valid to have `@foo:bar` and a `RequestOrigin` that isn't `bar`?

handlejoin.go
```diff
@@ -0,0 +1,278 @@
...
+	RequestDestination spec.ServerName
+	LocalServerName    spec.ServerName
```
kegsay, line 20: Isn't this request origin?

handlejoin.go
```diff
@@ -0,0 +1,278 @@
...
+	LocalServerInRoom  bool
```
kegsay, line 21: This doesn't make sense to me. It's never valid for this to be `false`, so why have this at all? I don't really know what the point of this flag is if it cannot handle both true/false? You can see this in the tests which are constantly forced into doing `LocalServerInRoom: true,`

kegsay (commented):
New API needs more comments, and I think some fields are probably unnecessary, but otherwise this is looking good. I'm happy to handle send_join in another PR.

handlejoin.go
kegsay: Because this is the receiving side

handlejoin.go
kegsay: Because this is the receiving side

handlejoin.go
kegsay: Because this is the receiving side

kegsay (approved):
Please comment the fields in `HandleMakeJoinInput`. Otherwise LGTM!

PR #383, opened 2023-05-16
Remove methods returning/using `EventReference`

S7evinK:
Also removes `CacheCost` which is now implemented in Dendrite.

### DIFF
--- a/event_builder.go
+++ b/event_builder.go
@@ -85,7 +85,7 @@ func toEventReference(data any) []eventReference {
 		for _, eventID := range evs {
 			newEvents = append(newEvents, eventReference{
 				EventID:     eventID,
-				EventSHA256: base64FromEventID(eventID),
+				EventSHA256: eventHashFromEventID(eventID),
 			})
 		}
 		return newEvents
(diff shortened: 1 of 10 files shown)


kegsay (approved):

kegsay (approved):

event_builder.go
```diff
@@ -112,22 +154,18 @@ func (eb *EventBuilder) Build(
...
-		// event IDs as a []string, rather than full-blown []EventReference.
+		// event IDs as a []string, rather than full-blown []eventReference.
 		// Since gomatrixserverlib otherwise deals with EventReferences,
 		// take the event IDs out of these and replace the prev_events and
 		// auth_events with those new arrays.
 		switch prevEvents := eventStruct.PrevEvents.(type) {
 		case []string:
 			eventStruct.PrevEvents = prevEvents
-		case []EventReference:
+		case []eventReference:
```
kegsay, line 168: Do we need this? I thought EventFormatV2 == strings only, so this case should never be hit?

event_builder.go
```diff
@@ -196,3 +234,17 @@ func (eb *EventBuilder) Build(
 
 	return res, nil
 }
+
+// Base64FromEventID returns, if possible, the base64bytes representation
+// of the given eventID. Returns an empty spec.Base64Bytes if an error occurs decoding.
+func base64FromEventID(eventID string) spec.Base64Bytes {
```
kegsay, line 240: Call this `eventHashFromEventID` as that is more accurate.

kegsay (approved):

PR #384, opened 2023-05-19
Move SendJoin logic from Dendrite

devonh:
Moves the core matrix logic for handling the send_join endpoint over from dendrite.


kegsay (approved):
Looks great

PR #387, opened 2023-05-25
Move `/send_leave` to GMSL

S7evinK:


handleleave_test.go
```diff
@@ -221,3 +222,177 @@ func TestHandleMakeLeave(t *testing.T) {
 		})
 	}
 }
+
+type dummyQuerier struct {
+	pdu PDU
```
kegsay, line 227: It would be good to assert the type/state key are correct when queried via `CurrentStateEvent`.

handleleave.go
```diff
@@ -98,3 +101,124 @@ func HandleMakeLeave(input HandleMakeLeaveInput) (*HandleMakeLeaveResponse, erro
 	}
 	return &makeLeaveResponse, nil
 }
+
+type CurrentStateQuerier interface {
+	CurrentStateEvent(ctx context.Context, roomID spec.RoomID, eventType string, stateKey string) (PDU, error)
+}
+
+// HandleSendLeave handles requests to `/send_leave
+// Returns the parsed event and an error.
```
kegsay, line 110:
```suggestion
// Returns the parsed event or an error.
```

handleleave.go
```diff
@@ -98,3 +101,124 @@ func HandleMakeLeave(input HandleMakeLeaveInput) (*HandleMakeLeaveResponse, erro
...
+func HandleSendLeave(ctx context.Context,
+	requestContent []byte,
+	origin spec.ServerName,
+	roomVersion RoomVersion,
+	eventID, roomID string,
+	querier CurrentStateQuerier,
+	verifier JSONVerifier,
+) (PDU, error) {
```
kegsay, line 118: Sidenote: this function and others like it will likely need to be modified to support pseudo IDs, as in many cases we pull out the state key and check that it is a user ID, and then check that the domain is correct on it. I would probably structure this as:
 - static function `HandleSendLeave` as it is in this PR
 - Pull out the room version immediately.
 - Call `roomVer.HandleSendLeave(args...)` to let room versions decide how to implement this.
 - Most room versions will call a private function which does the below code, checking user IDs etc.
 - But pseudo IDs will call a custom function which omits these checks.

Whilst the caller _could_ call `roomVer.HandleSendLeave` this then creates bad symmetry, as there are cases where you don't know the room version at this point (e.g invites).

kegsay (approved):

eventversion.go
```diff
@@ -431,6 +447,17 @@ func (v RoomVersionImpl) RedactEventJSON(eventJSON []byte) ([]byte, error) {
 	return v.redactionAlgorithm(eventJSON)
 }
 
+// HandleSendLeave handles requests to `/send_leave`
+func (v RoomVersionImpl) HandleSendLeave(ctx context.Context,
+	event PDU,
```
kegsay, line 452: This is problematic if `v != event.Version()`.

handleleave.go
```diff
@@ -106,12 +106,11 @@ type CurrentStateQuerier interface {
...
 }
 
-// HandleSendLeave handles requests to `/send_leave
-// Returns the parsed event and an error.
-func HandleSendLeave(ctx context.Context,
-	requestContent []byte,
+// handleSendLeave handles requests to `/send_leave
+// Returns the parsed event or an error.
+func handleSendLeave(ctx context.Context,
+	event PDU,
```
kegsay, line 112: The proposal I mentioned in https://github.com/matrix-org/gomatrixserverlib/pull/387#discussion_r1214123648 was to keep this function public, and to serve as the entry point for callers, because:

> Whilst the caller could call roomVer.HandleSendLeave this then creates bad symmetry, as there are cases where you don't know the room version at this point (e.g invites).

PR #388, opened 2023-05-30
Use functions instead of bools for `IRoomVersion`

S7evinK:
... where possible


kegsay (approved):
This is fine for now, as it's more into a shape we want than before. That being said, the functions that are here are not ideal and way too specific. We can fix that in another PR though.

PR #390, opened 2023-06-03
PDU Sender split

devonh:
Initial cut of splitting PDU Sender into SenderID & doing a UserID lookup where applicable.


eventauth.go
```diff
@@ -801,13 +795,17 @@ func (a *allowerContext) redactEventAllowed(event PDU) error {
 	// sender and the redacted event.
 	// We leave it up to the sending server to implement the additional checks
 	// to ensure that only events that should be redacted are redacted.
-	if senderDomain == redactDomain {
+	sender, err := event.UserID()
+	if err != nil {
+		return err
+	}
+	if string(sender.Domain()) == redactDomain {
```
kegsay, line 802: This check doesn't make sense in room versions 3+ because event IDs are hashes, so `redactDomain` will be nonsensical, which is why we return early on `:783`. As we are basing pseudo IDs on modern room versions, the fact this uses domains and user IDs is fine.

eventauth.go
```diff
@@ -859,35 +857,37 @@ func (e *eventAllower) commonChecks(event PDU) error {
...
 
-	sender := event.Sender()
 	stateKey := event.StateKey()
-
-	if err := e.create.UserIDAllowed(sender); err != nil {
+	userID, err := event.UserID()
+	if err != nil {
+		return err
+	}
+	if err := e.create.UserIDAllowed(userID); err != nil {
```
kegsay, line 865: AKA "can we federate with this user ID?" which is unenforcable for pseudo IDs, so we'd remove it.

eventauth.go
```diff
@@ -859,35 +857,37 @@ func (e *eventAllower) commonChecks(event PDU) error {
...
-			event.Sender(), senderLevel, eventLevel,
+			event.SenderID(), senderLevel, eventLevel,
 		)
 	}
 
 	// Check that all state_keys that begin with '@' are only updated by users
 	// with that ID.
 	if stateKey != nil && len(*stateKey) > 0 && (*stateKey)[0] == '@' {
-		if *stateKey != sender {
+		if *stateKey != event.SenderID() {
```
kegsay, line 887: This check is what I've spoken about before, which is a problem for pseudo IDs as they do NOT have `@` sigil, and yet will be used in cases where the events are only intended to be updated by users with the same ID. I don't think this is a strong reason to introduce a new sigil e.g `^` like MSC1228 suggests. We do however need to figure out a way to resolve this.

eventauth.go
```diff
@@ -960,10 +960,18 @@ func (m *membershipAllower) membershipAllowed(event PDU) error { // nolint: gocy
...
+	}
+	if err := m.create.UserIDAllowed(*sender); err != nil {
+		return err
+	}
+	target, err := spec.NewUserID(m.targetID, true)
+	if err != nil {
 		return err
 	}
-	if err := m.create.UserIDAllowed(m.targetID); err != nil {
+	if err := m.create.UserIDAllowed(*target); err != nil {
```
kegsay, line 974: AKA can we federate with this user.

kegsay (approved):
Many comments, but mostly just to point out that "this is correct, but will need to change when we work on pseudo IDs" mostly so we don't forget.

Outstanding Qs around:

> Check that all state_keys that begin with '@' are only updated by users with that ID.

kegsay (approved):

PR #393, opened 2023-06-07
Cleanup remaining statekey usage for senderIDs

devonh:


eventV2.go
```diff
@@ -210,9 +210,12 @@ func CheckFields(input PDU) error { // nolint: gocyclo
 		return err
 	}
 
-	// TODO: Use input.Version to check if this should be a senderKey or userID
-	if err := checkID(string(input.SenderID()), "user", '@'); err != nil {
-		return err
+	switch input.Version() {
+	case RoomVersionV1, RoomVersionV2, RoomVersionV3, RoomVersionV4, RoomVersionV5,
+		RoomVersionV6, RoomVersionV7, RoomVersionV8, RoomVersionV9, RoomVersionV10:
```
kegsay, line 215: Can we invert and default instead? E.g `case RoomVersionPseduoID:` is empty then `default:` contains the check? It's going to be a footgun to remember to keep this switch-case updated otherwise.

eventauth.go
```diff
@@ -487,7 +486,7 @@ func (a *allowerContext) aliasEventAllowed(event PDU) error {
 	// Check that event is a state event.
 	// Check that the state key matches the server sending this event.
 	// https://github.com/matrix-org/synapse/blob/v0.18.5/synapse/api/auth.py#L158
-	if !event.StateKeyEquals(string(sender.Domain())) {
+	if !event.StateKeyEquals(string(event.SenderID())) {
```
kegsay, line 489: This isn't correct for normal Matrix versions which have the state key set to the domain of the HS e.g `state_key: matrix.org`.

eventauth.go
```diff
@@ -1035,9 +1034,7 @@ func (m *membershipAllower) membershipAllowedSelfForRestrictedJoin() error {
 	// 'join_authorised_via_users_server' key, containing the user ID of a user
 	// in the room that should have a suitable power level to issue invites.
 	// If no such key is specified then we should reject the join.
-	if _, _, err := SplitID('@', m.newMember.AuthorisedVia); err != nil {
-		return errorf("the 'join_authorised_via_users_server' contains an invalid value %q", m.newMember.AuthorisedVia)
```
kegsay, line 1039: We can't remove this check for normal room versions...

eventcrypto.go
```diff
@@ -78,11 +78,18 @@ func VerifyEventSignatures(ctx context.Context, e PDU, verifier JSONVerifier, us
 
 		// For invites, the invited server should have signed the event.
 		if membership == spec.Invite {
-			_, serverName, err = SplitID('@', *e.StateKey())
-			if err != nil {
-				return fmt.Errorf("failed to split state key: %w", err)
+			switch e.Version() {
+			case RoomVersionV1, RoomVersionV2, RoomVersionV3, RoomVersionV4, RoomVersionV5,
+				RoomVersionV6, RoomVersionV7, RoomVersionV8, RoomVersionV9, RoomVersionV10:
```
kegsay, line 83: Same deal, please invert

eventcrypto.go
```diff
@@ -78,11 +78,18 @@ func VerifyEventSignatures(ctx context.Context, e PDU, verifier JSONVerifier, us
...
+				_, serverName, err = SplitID('@', *e.StateKey())
+				if err != nil {
+					return fmt.Errorf("failed to split state key: %w", err)
+				}
+				needed[serverName] = struct{}{}
+			default:
+				// TODO: (pseudoIDs) revisit this logic for event signing
```
kegsay, line 90: Specifically the invitee should have signed the event, ergo we need a /make_invite which will sign this.

kegsay (changes requested):

eventversion.go
```diff
@@ -48,16 +48,17 @@ type EventIDFormat int
...
+	RoomVersionV2        RoomVersion = "2"
+	RoomVersionV3        RoomVersion = "3"
+	RoomVersionV4        RoomVersion = "4"
+	RoomVersionV5        RoomVersion = "5"
+	RoomVersionV6        RoomVersion = "6"
+	RoomVersionV7        RoomVersion = "7"
+	RoomVersionV8        RoomVersion = "8"
+	RoomVersionV9        RoomVersion = "9"
+	RoomVersionV10       RoomVersion = "10"
+	RoomVersionPseudoIDs RoomVersion = "PseudoIDs"
```
S7evinK, line 61:
```suggestion
	RoomVersionPseudoIDs RoomVersion = "org.matrix.msc4014"
```
I guess?

PR #394, opened 2023-06-13
Add `MXIDMapping` struct and possibility to sign it

S7evinK:


eventcrypto.go
```diff
@@ -80,6 +80,14 @@ func VerifyEventSignatures(ctx context.Context, e PDU, verifier JSONVerifier, us
 			return fmt.Errorf("failed to get membership of membership event: %w", err)
 		}
 
+		// Validate the MXIDMapping is signed correctly
+		if verImpl.Version() == RoomVersionPseudoIDs {
```
kegsay, line 84: `VerifyEventSignatures` should be part of `RoomVersion` so we don't need conditionals like this.

kegsay (approved):

eventcrypto.go
kegsay: Doesn't have to be in this PR tho.

PR #397, opened 2023-06-20
Handle pseudo identity room federated joins

S7evinK:
Feels a little bit hacky, but allows federated joins for `org.matrix.msc4014` rooms.


performjoin.go
```diff
@@ -213,6 +237,40 @@ func PerformJoin(
 		}
 	}
 
+	// get the membership event of the room creator, so we can store the mxid_mapping
+	// TODO: better way?
+	if roomVersion == RoomVersionPseudoIDs {
```
devonh, line 242: interesting... should we be doing a similar thing with the respSendJoin state events to add mappings for all the currently joined users? (not just the creator)
seems like an expensive operation, but possibly necessary...

S7evinK: Good point, definitely makes sense to add that.

performjoin.go
```diff
@@ -22,9 +22,10 @@ type PerformJoinInput struct {
 	KeyID      KeyID              // Used to sign the join event
 	KeyRing    *KeyRing           // Used to verify the response from send_join
 
-	EventProvider   EventProvider        // Provides full events given a list of event IDs
-	UserIDQuerier   spec.UserIDForSender // Provides userID for a given senderID
-	SenderIDCreator spec.CreateSenderID  // Creates new senderID for this room
+	EventProvider             EventProvider                  // Provides full events given a list of event IDs
+	UserIDQuerier             spec.UserIDForSender           // Provides userID for a given senderID
+	SenderIDCreator           spec.CreateSenderID            // Creates new senderID for this room
```
devonh, line 27: This could use a better name. It's not just a creator. It will create one if none exists, but otherwise it will return the existing senderID.

devonh: And the comment should be updated as well

PR #400, opened 2023-07-06
Add event size checks similar to Synapse

S7evinK:


eventV2.go
```diff
@@ -190,22 +207,43 @@ func CheckFields(input PDU) error { // nolint: gocyclo
...
 
+	_, persistable := lenientByteLimitRoomVersions[input.Version()]
+
+	if l := len(input.Type()); l > maxIDLength {
+		return EventValidationError{
+			Code:        EventValidationTooLarge,
+			Message:     fmt.Sprintf("gomatrixserverlib: event type is too long, length %d bytes > maximum %d bytes", l, maxIDLength),
+			Persistable: persistable,
+		}
+	}
```
devonh, line 235: We should comment the reasoning behind the 2 checks like synapse does.

PR #404, opened 2023-07-20
Fix signatures on pseudoID invites

devonh:


performinvite_test.go
```diff
@@ -25,6 +25,14 @@ func CreateSenderID(ctx context.Context, userID spec.UserID, roomID spec.RoomID,
 	return spec.SenderID(userID.String()), key, nil
 }
 
+func CreateSenderIDPseudoIDs(ctx context.Context, userID spec.UserID, roomID spec.RoomID, roomVersion string) (spec.SenderID, ed25519.PrivateKey, error) {
+	_, key, err := ed25519.GenerateKey(nil)
+	if err != nil {
+		panic("failed generating ed25519 key")
```
S7evinK, line 31: Can we just return the error instead of panicking?

S7evinK (approved):

PR #410, opened 2023-08-15
Disable character range check for user IDs

neilalexander:
This check is not safe in existing room versions because Synapse hasn't enforced it properly either.

Related: matrix-org/matrix-spec#1506, matrix-org/matrix-spec-proposals#2828

Signed-off-by: Neil Alexander <git@neilalexander.dev>


neilalexander:
cc @S7evinK @devonh

kegsay (changes requested):
This is the hill I'm dying on.

https://github.com/matrix-org/matrix-spec/pull/1506#discussion_r1294550066

If the above clarification PR lands, then I'll admit defeat and merge this.

PR #415, opened 2023-09-21
Extend fed client interface for complement usage

devonh:


kegsay (approved):
LGTM, but surely we need a concrete impl of these functions somewhere in GMSL too?

devonh:
> LGTM, but surely we need a concrete impl of these functions somewhere in GMSL too?

They already exist, we just weren't exposing them in the interface since Dendrite must not use them.

PR #419, opened 2023-10-06
Fix possible state reset issue

S7evinK:
The spec says:
> **Iterative auth checks.** The iterative auth checks algorithm takes as input an initial room state and a sorted list of state events, and constructs a new room state by iterating through the event list and applying the state event to the room state if the state event is allowed by the [authorization rules](https://spec.matrix.org/v1.8/server-server-api#authorization-rules). If the state event is not allowed by the authorization rules, then the event is ignored. If a `(event_type, state_key)` key that is required for checking the authorization rules is not present in the state, then the appropriate state event from the event’s `auth_events` is used if the auth event is not rejected.

The part we didn't do, given the membership event may not be resolved (e.g. it is conflicted):
> If a `(event_type, state_key)` key that is required for checking the authorization rules is not present in the state, then the appropriate state event from the event’s `auth_events` is used if the auth event is not rejected.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -16,7 +16,7 @@ linters:
     - misspell
     - unparam
     - goimports
-    - goconst
+#    - goconst
     - unconvert
     - errcheck
     - interfacer
--- a/stateresolutionv2.go
+++ b/stateresolutionv2.go
@@ -450,8 +450,18 @@ func (r *stateResolverV2) authAndApplyEvents(events []PDU) {
 			_ = r.authProvider.AddEvent(event)
 		}
 		for _, needed := range needed.Member {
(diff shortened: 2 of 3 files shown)


devonh (approved):
Can you add a comment to the test outlining which prev_events are `wrong`?
It's impossible to tell what the event_ids are just from looking at the code.

Otherwise this looks like it should meet the rules outlined in the spec.

S7evinK:
There isn't anything wrong with the `prev_events` in this test. The problem is mainly that Dendrite is "wrongly" calculating the conflicted events. See [this](https://github.com/matrix-org/gomatrixserverlib/pull/419/files#diff-8625a2d36ba013326beaedf4bfe0da63a92f7edc7a59c210c562de1685d4158aR558-R561).

PR #421, opened 2023-10-24
Check if event is rejected before using it for auth

S7evinK:
To be even more spec compliant, this adds a `IsRejected` type/function when resolving state.
(Part 2 of https://github.com/matrix-org/gomatrixserverlib/pull/419)

### DIFF
--- a/stateresolutionv2.go
+++ b/stateresolutionv2.go
@@ -46,15 +46,20 @@ type stateResolverV2 struct {
 	resolvedMembers           map[spec.SenderID]PDU         // Resolved member events
 	resolvedOthers            map[StateKeyTuple]PDU         // Resolved other events
 	result                    []PDU                         // Final list of resolved events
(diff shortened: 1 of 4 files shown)


stateresolutionv2.go
```diff
@@ -46,15 +46,20 @@ type stateResolverV2 struct {
...
+
 // ResolveStateConflicts takes a list of state events with conflicting state
 // keys and works out which event should be used for each state event. This
 // function returns the resolved state, including unconflicted state events.
 func ResolveStateConflictsV2(
-	conflicted, unconflicted []PDU,
+	conflicted, unconflicted,
 	authEvents []PDU,
 	userIDForSender spec.UserIDForSender,
+	isRejected IsRejected,
```
kegsay, line 62: Please rename this to `isRejectedFn` everywhere. I got very confused and though this was a `bool` which doesn't make sense.

kegsay (approved):

PR #422, opened 2023-11-10
Avoid logging issues if we still could fetch keys

S7evinK:
To avoid confusion about e.g. being unable to fetch keys from matrix.org.

### DIFF
--- a/keyring.go
+++ b/keyring.go
@@ -499,7 +499,6 @@ func (d DirectKeyFetcher) FetcherName() string {
 func (d *DirectKeyFetcher) FetchKeys(
 	ctx context.Context, requests map[PublicKeyLookupRequest]spec.Timestamp,
 ) (map[PublicKeyLookupRequest]PublicKeyLookupResult, error) {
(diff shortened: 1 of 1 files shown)


keyring.go
```diff
@@ -290,6 +289,19 @@ func (k KeyRing) VerifyJSONs(ctx context.Context, requests []VerifyJSONRequest)
...
+	if len(keyRequests) > 0 {
+		requestedServers := make([]string, 0, len(keyRequests))
+		for reqs := range keyRequests {
+			requestedServers = append(requestedServers, string(reqs.ServerName))
+		}
+
+		logger.WithFields(logrus.Fields{
+			"servers":  requestedServers,
+			"fetchers": len(k.KeyFetchers),
+		}).Warn("failed to fetch keys for some servers")
```
kegsay, line 302: Without stating which servers this isn't very helpful.

S7evinK: `requestedServers` is the list of the servers we failed to fetch keys for.

kegsay: Ah, that wasn't clear. LGTM then!

kegsay (approved):

PR #424, opened 2023-12-08
Handle negative zero and exponential numbers in Canonical JSON verification

devonh:


devonh:
There are commented out test cases for other invalid canonical json that we don't currently detect.
In the future we should detect them, but for now I have left them out in order to get these simple and clear violations fixed.

json_test.go
```diff
@@ -78,9 +79,10 @@ func TestSortJSON(t *testing.T) {
 }
 
 func testCompactJSON(t *testing.T, input, want string) {
-	got := string(CompactJSON([]byte(input), nil))
+	bytes := CompactJSON([]byte(input), nil)
+	got := string(bytes)
 	if got != want {
-		t.Errorf("CompactJSON(%q):\n want: %q\n got: %q", input, want, got)
+		t.Errorf("CompactJSON(%q):\n want: %q\n got: %q\n bytes: % X", input, want, got, bytes)
```
S7evinK, line 85:
```suggestion
		t.Errorf("CompactJSON(%q):\n want: %q\n got: %q\n bytes: %x", input, want, got, bytes)
```
I assume? (with the recent lower-/uppercase thing?)

S7evinK (approved):

json_test.go
S7evinK: From https://pkg.go.dev/fmt
```
' '	(space) leave a space for elided sign in numbers (% d);
	put spaces between bytes printing strings or slices in hex (% x, % X)
```

So this is fine.

PR #430, opened 2024-01-15
Fix some event auth issues

S7evinK:


eventauth.go
```diff
@@ -1180,6 +1187,13 @@ func (m *membershipAllower) membershipAllowedSelf() error { // nolint: gocyclo
 		if m.oldMember.Membership == spec.Join {
 			return nil
 		}
+
+		// A user that is not in the room is allowed to join if the room
+		// join rules are "public".
+		if m.oldMember.Membership == spec.Leave && m.joinRule.JoinRule == spec.Public {
+			return nil
+		}
```
devonh, line 1195: Do we need the `m.oldMember.Membership == spec.Leave` portion of the check?
I think it would be more accurate just to check if the room is public. The only case to reject the join for a public room would be if the user was banned, and that is already handled before this switch block.

eventauth_test.go
```diff
@@ -1591,5 +1602,340 @@ func TestMembershipAllowed(t *testing.T) {
...
+			"sender": "@u1:a",
+			"room_id": "!r1:a",
+			"state_key": "@u1:a",
+			"event_id": "$e2:a",
+			"content": {"membership": "join"},
+			"unsigned": {
+				"not_allowed": "Sender not invited or joined"
+			}
+		}]
+	}`, RoomVersionV1)
```
devonh, line 1791: Room version 10 I presume?

eventauth.go
```diff
@@ -1162,15 +1169,15 @@ func (m *membershipAllower) membershipAllowedSelf() error { // nolint: gocyclo
...
 			if err := m.membershipAllowedSelfForRestrictedJoin(); err != nil {
 				return err
 			}
-		}
-		// A user that is not in the room is allowed to join if the room
-		// join rules are "public".
-		if m.oldMember.Membership == spec.Leave && m.joinRule.JoinRule == spec.Public {
-			return nil
+			// If, after validating restricted joins, the room is now "public", allow.
+			// This means that transitions from knock|invite|leave to join are allowed.
```
devonh, line 1177: Is it actually acceptable to go from knock to join? I thought that wasn't allowed as per https://spec.matrix.org/v1.8/diagrams/membership.png

devonh: Diagram is wrong, this is allowed :)

eventauth_test.go
S7evinK: Shouldn't matter for `join_rule` set to `invite`, there were no changes to this, if I saw that correctly.

eventauth.go
S7evinK: https://github.com/matrix-org/matrix-spec/pull/1709

eventauth.go
S7evinK: I'd rather be explicit here, as not to open up for potential bugs.

PR #431, opened 2024-01-16
Fix event auth for knocking

S7evinK:
Knocking is not allowed in `restricted` rooms.


kegsay (approved):
Seems really broken that we allow invite->knock transitions... Invites are more important than knocks as you can then transition to join from it, but you can't with knock.

The only time I guess it's useful is to revoke an invite of a knocked user, but I would have probably advocated for invite->leave and then make the knocker knock again, but oh well. 

Codewise LGTM

PR #436, opened 2024-07-08
Trim whitespace in X-Matrix header

### LINKED ISSUE #434: gomatrixserverlib rejects `Authorization: X-Matrix ...` headers that include optional whitespace around parameter pairs
The [spec](https://spec.matrix.org/v1.10/server-server-api/#authentication) states that the format of the `Authorization` header on the Federation API is given by [RFC 7235](https://datatracker.ietf.org/doc/html/rfc7235#section-2.1). Part of that specification allows for zero or more spaces or tabs around the commas in the parameter list. For example, the following should be a valid `Authorization` header:

```
Authorization: X-Matrix origin=foo , key="ed25519:1",  sig="sig",		destination="bar"
```

Yet upon looking at `gomatrixserverlib`'s source code, I believe it would reject this header:

https://github.com/matrix-org/gomatrixserverlib/blob/c2391f7113a5020d28d9fa5680ae025c20062126/fclient/request.go#L321-L351

This fact was omitted from the summary of RFC 7235 in the spec, which is likely why it didn't make it into `gomatrixserverlib` as well. That is now being corrected: https://github.com/matrix-org/matrix-spec/pull/1818

The code should be updated to allow whitespace in the parameter list in the header, such that any homeserver implementations following the spec would not have their federation requests rejected by homeservers that rely on gomatrixserverlib.

See https://github.com/matrix-org/matrix-spec/issues/1817 for more context.

S7evinK:
Fixes #434

### DIFF
--- a/fclient/request.go
+++ b/fclient/request.go
@@ -333,7 +333,7 @@ func ParseAuthorization(header string) (scheme string, origin, destination spec.
 			continue
 		}
 		name := strings.TrimSpace(pair[0])
-		value := strings.TrimSpace(strings.Trim(strings.TrimSpace(pair[1]), "\""))
+		value := strings.Trim(strings.TrimSpace(pair[1]), "\"")
 		if name == "origin" {
 			origin = spec.ServerName(value)
 		}
--- a/fclient/request.go
+++ b/fclient/request.go
@@ -332,8 +332,8 @@ func ParseAuthorization(header string) (scheme string, origin, destination spec.
 		if len(pair) != 2 {
 			continue
 		}
-		name := pair[0]
-		value := strings.Trim(pair[1], "\"")
+		name := strings.TrimSpace(pair[0])
+		value := strings.TrimSpace(strings.Trim(strings.TrimSpace(pair[1]), "\""))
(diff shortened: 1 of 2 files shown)


fclient/request.go
```diff
@@ -332,8 +332,8 @@ func ParseAuthorization(header string) (scheme string, origin, destination spec.
 		if len(pair) != 2 {
 			continue
 		}
-		name := pair[0]
-		value := strings.Trim(pair[1], "\"")
+		name := strings.TrimSpace(pair[0])
+		value := strings.TrimSpace(strings.Trim(strings.TrimSpace(pair[1]), "\""))
```
devonh, line 336: Why does this call `TrimSpace` twice? Is once not enough in this case?

S7evinK: Since we're splitting the line on `,` and `=`, we may end up with something like `destination	=  "bar"  `. This means the inner `TrimSpace` results in `"bar"`, then we `Trim` the `"`, resulting in `bar` - and while writing this, yes, one `TrimSpace` is enough.

PR #437, opened 2024-07-27
Add support for getting authenticated federation media

S7evinK:


fclient/federationclient.go
```diff
@@ -736,3 +738,31 @@ func (ac *federationClient) RoomHierarchy(
 	}
 	return
 }
+
```
HenrikSolver, line 741: I think documentation in the same spirit as the other methods would be appropriate.

HenrikSolver (commented):
To my understanding this looks fine.

fclient/federationclient.go
S7evinK: Fair, and good to document that the caller of this method is responsible to close the returned response body, if any.

HenrikSolver (approved):
LGTM

S7evinK:
Thanks for the review! :)

Issue #440, opened 2024-10-29
LookupWellKnown fails to parse well-known if response spans multiple tcp segments


richvdh:
`LookupWellKnown` contains this code:

```go
	bodyBuffer := make([]byte, WellKnownMaxSize)
	limitedReader := &io.LimitedReader{
		R: resp.Body,
		N: WellKnownMaxSize,
	}
	n, err := limitedReader.Read(bodyBuffer)
	if err != nil && err != io.EOF {
		return nil, err
	}
	body := bodyBuffer[:n]
```

That will only read part of the response if the response spans multiple tcp segments.

I think this is equivalent, but works correctly:

```go
	body, err := io.ReadAll(&io.LimitedReader{
		R: resp.Body,
		N: WellKnownMaxSize,
	})
	if err != nil {
		return nil, err
	}
```

PR #441, opened 2024-10-29
Fix `LookupWellKnown` to read full response

richvdh:
Fix a bug in `LookupWellKnown` which would cause it only to read part of the HTTP response body if it spanned multiple TCP segments.

Fixes: https://github.com/matrix-org/gomatrixserverlib/issues/440


S7evinK (approved):
Hu, interesting. Wondering why that didn't bite us earlier.

PR #444, opened 2024-12-21
Add Equal method for CrossSigningKey

S7evinK:
Playing around with Copilot, code is mostly generated (but reviewed)

Also fixes CI by using the correct repository for Dendrite.

### DIFF
--- a/fclient/crosssigning.go
+++ b/fclient/crosssigning.go
@@ -15,8 +15,8 @@
 package fclient
 
 import (
+	"bytes"
 	"encoding/json"
-
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/gomatrixserverlib/spec"
 	"github.com/tidwall/gjson"
@@ -45,6 +45,49 @@ type CrossSigningKey struct {
 }
 
 func (s *CrossSigningKey) isCrossSigningBody() {} // implements CrossSigningBody
+
+func (s *CrossSigningKey) Equal(other *CrossSigningKey) bool {
+	if s == nil || other == nil {
+		return false
+	}
+	if s.UserID != other.UserID {
+		return false
+	}
+	if len(s.Usage) != len(other.Usage) {
+		return false
+	}
+	for i := range s.Usage {
+		if s.Usage[i] != other.Usage[i] {
+			return false
+		}
+	}
+	if len(s.Keys) != len(other.Keys) {
+		return false
+	}
+	for k, v := range s.Keys {
+		if !bytes.Equal(other.Keys[k], v) {
+			return false
+		}
+	}
+	if len(s.Signatures) != len(other.Signatures) {
+		return false
+	}
+	for k, v := range s.Signatures {
+		otherV, ok := other.Signatures[k]
+		if !ok {
+			return false
+		}
+		if len(v) != len(otherV) {
+			return false
+		}
+		for k2, v2 := range v {
+			if !bytes.Equal(otherV[k2], v2) {
+				return false
+			}
+		}
+	}
+	return true
+}
 
 type CrossSigningBody interface {
 	isCrossSigningBody()
--- a/fclient/crosssigning.go
+++ b/fclient/crosssigning.go
@@ -17,6 +17,7 @@ package fclient
 import (
 	"bytes"
 	"encoding/json"
+	"slices"
 
 	"github.com/matrix-org/gomatrixserverlib"
 	"github.com/matrix-org/gomatrixserverlib/spec"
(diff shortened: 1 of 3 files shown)


fclient/crosssigning.go
```diff
@@ -46,6 +47,49 @@ type CrossSigningKey struct {
 
 func (s *CrossSigningKey) isCrossSigningBody() {} // implements CrossSigningBody
 
+func (s *CrossSigningKey) Equal(other *CrossSigningKey) bool {
```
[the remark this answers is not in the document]
S7evinK: This method is does not allocate at all, while `reflect.DeepEqual` does. (not that it matters much on that endpoint)

```bash
> go test -bench=Equal -run=^# -benchmem -count=10
goos: linux
goarch: amd64
pkg: github.com/matrix-org/gomatrixserverlib/fclient
cpu: 12th Gen Intel(R) Core(TM) i5-12500H
BenchmarkReflectDeepEqual-16    	  468159	      2381 ns/op	     288 B/op	      14 allocs/op
BenchmarkReflectDeepEqual-16    	  460086	      2376 ns/op	     288 B/op	      14 allocs/op
BenchmarkReflectDeepEqual-16    	  494133	      2333 ns/op	     288 B/op	      14 allocs/op
BenchmarkReflectDeepEqual-16    	  496840	      2338 ns/op	     288 B/op	      14 allocs/op
BenchmarkReflectDeepEqual-16    	  481978	      2351 ns/op	     288 B/op	      14 allocs/op
BenchmarkReflectDeepEqual-16    	  461346	      2327 ns/op	     288 B/op	      14 allocs/op
BenchmarkReflectDeepEqual-16    	  511146	      2390 ns/op	     288 B/op	      14 allocs/op
BenchmarkReflectDeepEqual-16    	  501966	      2443 ns/op	     288 B/op	      14 allocs/op
BenchmarkReflectDeepEqual-16    	  503883	      2345 ns/op	     288 B/op	      14 allocs/op
BenchmarkReflectDeepEqual-16    	  503308	      2415 ns/op	     288 B/op	      14 allocs/op
BenchmarkEqual-16               	 5371452	       214.8 ns/op	       0 B/op	       0 allocs/op
BenchmarkEqual-16               	 5578748	       227.6 ns/op	       0 B/op	       0 allocs/op
BenchmarkEqual-16               	 5169262	       230.6 ns/op	       0 B/op	       0 allocs/op
BenchmarkEqual-16               	 5215892	       234.2 ns/op	       0 B/op	       0 allocs/op
BenchmarkEqual-16               	 5207256	       232.0 ns/op	       0 B/op	       0 allocs/op
BenchmarkEqual-16               	 5564870	       244.5 ns/op	       0 B/op	       0 allocs/op
BenchmarkEqual-16               	 4907815	       229.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkEqual-16               	 5178279	       233.5 ns/op	       0 B/op	       0 allocs/op
BenchmarkEqual-16               	 4840867	       247.9 ns/op	       0 B/op	       0 allocs/op
BenchmarkEqual-16               	 5193268	       215.0 ns/op	       0 B/op	       0 allocs/op
```

fclient/crosssigning.go
```diff
@@ -46,6 +47,49 @@ type CrossSigningKey struct {
...
+		return false
+	}
+	if s.UserID != other.UserID {
+		return false
+	}
+	if len(s.Usage) != len(other.Usage) {
+		return false
+	}
+	for i := range s.Usage {
+		if s.Usage[i] != other.Usage[i] {
```
poljar, line 61: Is this slice guaranteed to be ordered? If not you might get some false negatives.

fclient/crosssigning.go
poljar, line 50: It would be nice to document what the comparison is taking into account, i.e. [there can be different equality implementations](https://github.com/matrix-org/matrix-rust-sdk/blob/c50358366fe7f25606cb5494071391c39f74f43a/crates/matrix-sdk-crypto/src/types/cross_signing/mod.rs#L35-L45) depending on the use-case. I presume here we care about the fact that a new object needs to be stored in the DB and we test for equality of all fields.

fclient/crosssigning.go
S7evinK: Fair point, but does `usage` currently even contain more than one value?

fclient/crosssigning.go
S7evinK: Exactly. The `Equal` method here checks all fields and receives the `MasterKey`, `SelfSigningKeys` and `UserSigningKeys` from Dendrite, see [this](https://github.com/element-hq/dendrite/pull/3471/files#diff-a9dcd4fed4827bd67f59b29c2ee48f7e819d5f97c20d7d91423e4f7f6d8cdb36R136-R150) (which should be much like https://github.com/element-hq/synapse/pull/16943

fclient/crosssigning.go
poljar: Per spec no, but you want to be forwards compatible, otherwise you might need to prevent such cross signing keys from being uploaded.

S7evinK: Fixed in https://github.com/matrix-org/gomatrixserverlib/pull/444/commits/30cbdb4f140c61ae8a5c233bbe869aeb61610d27

PR #451, opened 2025-04-03
Pass RoundTripper to well known resolution

gaelgatelement:


gaelgatelement:
Cancelling after [discussing it in dendrite-devs room.](https://matrix.to/#/!BIDAeUqYWNCjRLhRdj:matrix.org/$R01ifSU6tYw45KAKtdHayU_WTyL5UCwwb4hygfbQvzg?via=matrix.org&via=element.io&via=envs.net)

PR #455, opened 2025-06-06
Add a way to set custom room version impls

kegsay:
We have this lovely room version abstraction, but... no one can actually use it without getting it into this map. Let's fix that.


eventversion.go
```diff
@@ -416,6 +418,14 @@ func StableRoomVersions() map[RoomVersion]IRoomVersion {
 	return versions
 }
 
+// SetRoomVersion sets a room version implementation so it is recognised by gomatrixserverlib.
+// This is useful when you are testing custom room versions which gomatrixserverlib may be unaware of,
+// e.g for Complement usage. Full room version impls should be defined in gomatrixserverlib, but
+// partial impls for testing can be set here.
+func SetRoomVersion(ver IRoomVersion) {
+	roomVersionMeta[ver.Version()] = ver
+}
```
devonh, line 427: Is there a small test we can add for this? Otherwise I can see it getting blown away accidentally since it's unused internally.

devonh (approved):
Other than a potential test addition, this looks perfectly reasonable to me.

PR #456, opened 2025-06-18
Fix an issue where we would send an invalid `/user/keys/query` request

S7evinK:
Discovered while checking Grafana and wondering why there was one server being hit over and over again.
Turns out that we were sending 

```json
{
  "device_keys": {
    "@alice:example.com": null
  }
}
```
instead of an empty list.

This should fix it, even though the server in question still seems to return an invalid response.

### Pull Request Checklist


S7evinK:
As the commit message suggests, tests are generated by Copilot.

PR #462, opened 2025-10-01
Add sticky event parsing

turt2live:
MSC: https://github.com/matrix-org/matrix-spec-proposals/pull/4354

### DIFF
--- a/eventV1.go
+++ b/eventV1.go
@@ -4,12 +4,17 @@ import (
 	"bytes"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"github.com/matrix-org/gomatrixserverlib/spec"
 	"github.com/tidwall/gjson"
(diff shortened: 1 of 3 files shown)


eventV1.go
```diff
@@ -21,6 +26,9 @@ type eventV1 struct {
 	EventIDRaw string           `json:"event_id,omitempty"`
 	PrevEvents []eventReference `json:"prev_events"`
 	AuthEvents []eventReference `json:"auth_events"`
+
+	UnstableSticky stickyEventData `json:"msc4354_sticky,omitempty"`
+	StableSticky   stickyEventData `json:"sticky,omitempty"`
```
devonh: Should this be applied to `eventV1`?
At a glance, I would think this only applies to `eventV3` or newer. But I haven't been following gmsl or that MSC so maybe I'm wrong.

PR #465, opened 2026-04-17
feat: add CreatorInCreateEvent to help with default room v11

FrenchGithubUser:
related to https://github.com/matrix-org/complement/pull/858

### Pull Request Checklist

Signed-off-by: `Thomas Traineau <t.traineau@famedly.com>`

### DIFF
--- a/eventversion.go
+++ b/eventversion.go
@@ -40,6 +40,9 @@ type IRoomVersion interface {
 	PrivilegedCreators() bool
 	StateDAGs() bool
 	CreatorInCreateEvent() bool
+	// StrictEventByteLimits returns true if this room version enforces field
+	// length limits in bytes rather than Unicode codepoints (introduced in v11 for synapse).
+	StrictEventByteLimits() bool
 }
 
 type KnownRoomVersionFunc func(RoomVersion) bool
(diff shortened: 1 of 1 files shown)


eventversion.go
```diff
@@ -39,6 +39,10 @@ type IRoomVersion interface {
 	DomainlessRoomIDs() bool
 	PrivilegedCreators() bool
 	StateDAGs() bool
+	CreatorInCreateEvent() bool
```
reivilibre, line 46: I know the others don't have this (boo!) but please could you add the doc comment above this, assuming that's what is normal to do in Go to make sure LSPs/IDEs can show the information on hover.

eventversion.go
```diff
@@ -110,6 +114,7 @@ var roomVersionMeta = map[RoomVersion]IRoomVersion{
 		checkKnockingAllowedFunc:               disallowKnocking,
 		checkRestrictedJoinAllowedFunc:         disallowRestrictedJoins,
 		checkCreateEvent:                       checkCreateEventV1,
+		creatorInCreateEvent:                   true,
```
reivilibre, line 121: I think most of these should be false, if I understand it right?

eventversion.go
```diff
@@ -529,6 +549,10 @@ type RoomVersionImpl struct {
 	domainlessRoomID bool
 	// creators have infinite PL
 	privilegedCreators bool
```
reivilibre, line 555: is `privilegedCreators` the same or different to `creatorInCreateEvent`?

I'd appreciate a clarification, probably worth writing down

reivilibre (changes requested):
I guess I don't understand what this is (maybe the comments can be a bit more enlightening, it probably needs too much context from the spec changes over the years)

eventversion.go
FrenchGithubUser: this is false from room v11 and above

eventversion.go
FrenchGithubUser: yes those are different, commenting about it in the next commit

FrenchGithubUser:
sorry about the missing comments. Let me know if there are further questions

kegsay (approved):

PR #466, opened 2026-06-22
Replace archived matrix-org/gomatrix dependency

Johennes:
This switches from https://github.com/matrix-org/gomatrix (which is archived) to https://github.com/mautrix/go.

### Pull Request Checklist

* [X] Pull request includes a [sign off](https://github.com/matrix-org/dendrite/blob/master/docs/CONTRIBUTING.md#sign-off)


Johennes:
Looks like this won't be as straightforward as I thought because apparently dendrite would require changes, too.

kegsay (approved):
LGTM, though I suspect this is a breaking change as users of this library will type assert gomatrix errors.

Johennes:
Yeah. 😕 Not sure how to avoid that though?

S7evinK:
iirc, we tried to never break Dendrite but didn't really care about gmsl.
If Dendrite works with this change, LGTM. Given the unit tests are failing, I'd be cautious.

.github/workflows/tests.yml
```diff
@@ -87,7 +87,10 @@ jobs:
     steps:
       - uses: actions/checkout@v3
         with:
-          repository: "element-hq/dendrite"
+          # TODO: Switch back to element-hq/dendrite once
+          # https://github.com/element-hq/dendrite/pull/3708 is merged
+          repository: "Johennes/dendrite"
+          ref: "johannes/gomatrix"
```
Johennes, line 93: This is for testing and demonstration purposes only and needs to be removed before merging.

Johennes:
I don't see a way to not break dendrite due to the type change. I've now downgraded mautrix/go to the same version used in dendrite and opened https://github.com/element-hq/dendrite/pull/3708 to replace matrix-org/gomatrix in dendrite, too.

PR #467, opened 2026-06-30
Ensure checking `RoomID`s works for room version 12

jason-famedly:
### Pull Request Checklist

Signed-off-by: `Jason Little <j.little@famedly.com>`

### DIFF
--- a/event.go
+++ b/event.go
@@ -65,9 +65,9 @@ const (
 )
 
 func checkID(id, kind string, sigil byte) (err error) {
-	if kind != "room" {
+	if kind == "room" {
 		// RoomIDs should always be an opaque string, so only do this check for other kinds of ID
-		if _, err = domainFromID(id); err != nil {
+		if _, err = spec.ParseAndValidateRoomID(id); err != nil {
 			return
 		}
 	}
--- a/event.go
+++ b/event.go
@@ -65,6 +65,12 @@ const (
 )
 
 func checkID(id, kind string, sigil byte) (err error) {
+	if kind != "room" {
+		// RoomIDs should always be an opaque string, so only do this check for other kinds of ID
+		if _, err = domainFromID(id); err != nil {
+			return
+		}
+	}
 	if id[0] != sigil {
 		err = fmt.Errorf(
 			"gomatrixserverlib: invalid %s ID, wanted first byte to be '%c' got '%c'",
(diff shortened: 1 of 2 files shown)


jason-famedly:
Actually, you know what? This is used in a couple different places for other sigil checks. Let me re-evaluate. I was to hasty.

jason-famedly:
> Actually, you know what? This is used in a couple different places for other sigil checks. Let me re-evaluate. I was to hasty.

There, that is more inclusive of uses other `kind == room`. It probably would have been fine, but this is probably better

event.go
```diff
@@ -65,8 +65,11 @@ const (
 )
 
 func checkID(id, kind string, sigil byte) (err error) {
-	if _, err = domainFromID(id); err != nil {
-		return
+	if kind != "room" {
+		// RoomIDs should always be an opaque string, so only do this check for other kinds of ID
```
kegsay: This isn't true though for rooms <v12.

It would be better to assert that the room ID either:
 - has a domain OR
 - is exactly 43 characters of base 64 (we do a similar check [in Synapse](https://github.com/element-hq/synapse/blob/develop/synapse/types/__init__.py#L431))

jason-famedly: Perhaps it would be better to just use [`parseAndValidateRoomID()`](https://github.com/matrix-org/gomatrixserverlib/blob/c9c468727353c4835523349dc7c202e5de9a2e37/spec/roomid.go#L44) and discard the returned `RoomID`?

jason-famedly: Come to think of it, the original premise of was discussed in the Complement room was about just removing the check for "TrustedJson" entirely. Perhaps should just go down that road instead of any of this?

kegsay: It's not about removing the check, it's about making the check work for v12 rooms. Reusing `parseAndValidateRoomID()` makes sense.

jason-famedly: Awesome, thanks!

nico-famedly: The spec does say that room ids were always opaque strings and should always have been treated as such, at least by clients: https://spec.matrix.org/v1.17/appendices/#room-ids

> Room IDs are case-sensitive and not meant to be human-readable. They are intended to be used as fully opaque strings by clients, even when a domain component is present.

But I guess this makes sense here, because servers are allowed to peek into it?

kegsay (approved):

